PR #381, opened 2015-01-09
-f option works

Sentient07:
Added the target to the target list when there is an IntegrityError(already the target exists in the DB)
Error messages won't be saved to the DB when -f option is used

### DIFF
--- a/framework/config/config.py
+++ b/framework/config/config.py
@@ -194,11 +194,13 @@ class Config(BaseComponent, ConfigInterface):
         scope = self.prepare_url_scope(options['Scope'], options['PluginGroup'])
         added_targets = []
         for target in scope:
+            added_targets.append(target)
             try:
                 self.target.AddTarget(target)
                 added_targets.append(target)
             except DBIntegrityException:
                 logging.warning("%s already exists in DB" % target)
+                added_targets.append(target)
             except UnresolvableTargetException as e:
                 logging.error("%s" % e.parameter)
         return(added_targets)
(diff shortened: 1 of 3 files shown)


DePierre:
**IMHO** that is not the way to go for a couple of reasons:

+ @alessandrofg has recently merge his ``re-architecture`` project but instead of using the ``.get_component`` you wrote a static variable for ``ErrorHandler`` and imported ``error_handler`` in ``owtf.py`` to update that variable
+ in ``error_handler.py`` line [107](https://github.com/owtf/owtf/pull/381/files#diff-b1fa914b814d6703b86247019e470003R107) you do not test if ``cl_arg`` contains ``"Force_Overwrite"``
+ the [change](https://github.com/owtf/owtf/pull/381/files#diff-9700eadd491a8f09de8d45e234631ef1R202) in ``config.py`` makes the code repetitive
+ (plus some minor pep8 syntax errors but that is not my point)

My point is that, **IMHO**, the fix should only be localized to ``config.py`` and should not touch ``error_handler.py`` nor ``owtf.py``. Your hack with static variable+method is what it is, a hack, not a fix.

@7a @tunnelshade a thought?

7a:
Hi Ramana,

Many things changed with the new architecture and Tao is right.
Could you try to fix that pull and send it again please?

Thanks!

DePierre:
As discussed and seen with @Sentient07 a new pull request will be opened for a better fix.

PR #384, opened 2015-01-24
Fixed issue 359 , 119 ,371

Ahiknsr:


DePierre:
Hi @Ahiknsr and thank you for your contribution.

I did not review your patch in depth but could you make your script PEP8 compliant please?

As soon as I have a full internet connection back, I will review your pr.

DePierre:
@Ahiknsr Hi,

I checked your update but that's not pep8 compliant, maybe according to some scripts but not according to the [guideline itself](https://www.python.org/dev/peps/pep-0008/). For instance:

+ The comments should be above the line and not below
+ The import should follow the order ``standard imports -> third party imports -> local imports``
(Also could you please use ``from __future__ import print_function`` in order to move towards python3 compatibility?)
+ 2 empty lines when dealing with a 0-level indentation.

I checked the script itself and there are some lines that bother me:

+ Why when retrieving the modules do you create a ``list``, then cast it into a ``set`` and finally cast it again into a ``list``? Can't you use a ``set`` the whole time?
+ Why don't you use ``with open() as f:`` instead of having ``open`` and ``close`` when writing into the file?
+ Why don't you create two methods for better readability? One for parsing the online file and one for saving the list?

7a:
@Ahiknsr OWTF was a project initially started to learn python, please take
depierre's suggestions positively and keep rocking! :)

Ahiknsr:
Thanks for the suggestions @DePierre , @7a  .I am modifying the script according to the suggestions made by @DePierre :)

dictionaries/drupalmodulesextractor.py
```diff
@@ -0,0 +1,66 @@
...
+
+from __future__ import print_function
+import urllib
+try:
+    from bs4 import BeautifulSoup
+except:
+    print("you need BeautifulSoup module to use this script")
+
+
+DRUPAL_MODULES_LIST= 'http://ftp.drupal.org/files/projects/'
```
DePierre: A whitespace is missing before the '='

dictionaries/drupalmodulesextractor.py
```diff
@@ -0,0 +1,66 @@
...
+OUTPUT = 'drupalmodules.txt'
+
+
+def module_extractor(links):
+    """This function will extract the module names from links"""
+    print("extracting module names")
+    modules = set()
+    for link in links:
+        if '-' in link['href']:
+            mod = link['href'].split('-')[0]
```
DePierre: L.52 and L.53 should be squashed into one line since the temporary ``mod`` variable is not used.

dictionaries/drupalmodulesextractor.py
```diff
@@ -0,0 +1,66 @@
...
+            modules.add(mod)
+    return modules
+
+
+# This page contains list of all the modules
```
DePierre: This comment should be moved right above ``DRUPAL_MODULES_LIST``.

DePierre:
@Ahiknsr I checked your updates and it's better now, thank you. I commented a couple of few things that should still be updated so please read them and update your script accordingly ;)

DePierre:
@Ahiknsr Hi again, I checked your update and it looks OK enough to be merged (we will update the print outputs to be more user-friendly).

One last thing is now to invoke your script from the installation process. Because for now, your script is not invoked anywhere. @7a will give you more details about that, ok?

dictionaries/drupalmodulesextractor.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+print("fetching the web page")
+html = urllib.urlopen(DRUPAL_MODULES_LIST)
+soup = BeautifulSoup(html)
+links = soup.findAll('a', href=True)
+modules = module_extractor(links)
+with open(OUTPUT, 'w') as f:
+    for module in modules:
+        f.write(module+'\n')
```
DePierre: There should be whitespaces surrounding the '+' sign.

DePierre:
@Ahiknsr I am sorry but we will not merge a pull request that fixes several bugs at once for a simple reason: Keep track of everything.

Let's say your pr introduce a bug, if we want to track down the origin of the bug, we will find the merge of your pull request that touches different files at once, fixing different bugs at once and is therefore more difficult to understand.

The best is to create a pull request per fixed bug. I hope you understand. So what you can do is to create a new branch for each bug, cherry-pick the specific commits and create a pull request with each branch.

Please update your pull requests accordingly, thank you.

Issue #119, opened 2013-12-23
Explore CMS-Explorer dictionary alternatives for best results


7a:
This is an issue to explore an alternative automatic mechanism to produce high quality CMS-specific dictionaries using a better, more up-to-date alternative source than CMS-Explorer (current source).

When OWTF is installed, we do a merge of various useful dictionaries for bruteforcing purporses. We should try to make sure that these dictionaries are the best for the task at hand.

The problem here is that it looks like CMS-Explorer has not been updated since 2010:

https://code.google.com/p/cms-explorer/downloads/list

Possibly, because of that, CMS-Explorer is getting 404 errors from drupal URLs that do no longer exist and this means that. Our dictionaries are perhaps not really up-to-date.

Note the following lines from below:
404 error getting http://drupalcode.org/viewvc/drupal/contributions/themes/
404 error getting http://drupalcode.org/viewvc/drupal/contributions/modules/

Full output courtesy of Marios Kourtesis (thanks!):

2013-12-10 16:39:35 (688 KB/s) - `cms-explorer-1.0.tar.bz2' saved [176245/176245]

cms-explorer-1.0/
cms-explorer-1.0/LW2.pm
cms-explorer-1.0/LICENSE.txt
cms-explorer-1.0/wp_themes.txt
cms-explorer-1.0/cms-explorer.pl
cms-explorer-1.0/joomla_plugins.txt
cms-explorer-1.0/joomla_themes.txt
cms-explorer-1.0/drupal_themes.txt
cms-explorer-1.0/wp_plugins.txt
cms-explorer-1.0/drupal_plugins.txt
cms-explorer-1.0/README.txt
[*] Going into directory: /root/owtf/dictionaries/cms-explorer/cms-explorer-1.0
[*] Updating cms-explorer.pl dictionaries..
*****************************************************************
WARNING: No osvdb.org API key defined, searches will be disabled.
*****************************************************************
404 error getting http://drupalcode.org/viewvc/drupal/contributions/themes/
drupal_themes.txt updated with 12 entries
wp_themes.txt updated with 7653 entries
wp_plugins.txt updated with 39608 entries
404 error getting http://drupalcode.org/viewvc/drupal/contributions/modules/
drupal_plugins.txt updated with 50 entries
[*] Copying updated dictionaries from /root/owtf/dictionaries/cms-explorer/cms-explorer-1.0 to /root/owtf/dictionaries/restricted/cms
[...]

PR #389, opened 2015-01-30
fix for issue #119 

Ahiknsr:


dictionaries/drupalmodulesextractor.py
```diff
@@ -0,0 +1,65 @@
...
+
+This is used for generating a dictionary of drupal modules
+
+"""
+
+from __future__ import print_function
+import urllib
+try:
+    from bs4 import BeautifulSoup
+except:
```
tunnelshade: What 'except'. It must listen for particular exception

Issue #379, opened 2014-12-22
[develop] -f does not work


DePierre:
OWTF provides a ``-f`` option that allows the user to force some plugins to re-run and override their old results.

This ``-f`` option does not seem to work.

```bash
$ python2 owtf.py --nowebui -o OWTF-WVS-001 http://127.0.0.1
 _____ _ _ _ _____ _____
|     | | | |_   _|   __|
|  |  | | | | | | |   __|
|_____|_____| |_| |__|

        @owtfp
    http://owtf.org
    
[-] Defaulting Plugin Group to 'web' based on list of plugins supplied
[*] OWTF Version: 1.0.1, Release: LionHeart (Beta) 
[-] Loading framework please wait..
[-] The IP address for 127.0.0.1 is: '127.0.0.1'
[*] 127.0.0.1:8008 <-- HTTP(S) Proxy to which requests can be directed
[-] Work assigned to Worker-1 with pid 28214
[-] __________ 1 - Target: http://127.0.0.1 -> Plugin: Arachni Unauthenticated (web/active) __________
# [omitted]
[-] All jobs have been done. Exiting.
[-] Stopping inbound proxy processes and cleaning up, Please wait!

$ python2 owtf.py --nowebui -f -o OWTF-WVS-001 http://127.0.0.1
 _____ _ _ _ _____ _____
|     | | | |_   _|   __|
|  |  | | | | | | |   __|
|_____|_____| |_| |__|

        @owtfp
    http://owtf.org
    
[-] Defaulting Plugin Group to 'web' based on list of plugins supplied
[*] OWTF Version: 1.0.1, Release: LionHeart (Beta) 
[-] Loading framework please wait..
[*] http://127.0.0.1 already exists in DB
[*] 127.0.0.1:8008 <-- HTTP(S) Proxy to which requests can be directed
[-] All jobs have been done. Exiting.
[-] Stopping inbound proxy processes and cleaning up, Please wait!
```

A fix for this issue might be related to the one for #378.

Issue #378, opened 2014-12-16
[develop] Can't run OWTF more than once against the same target


DePierre:
Using the CLI, OWTF won't run more than once against the same target.
Instead, it logs that the target already exists and exits, even if the user runs a different plugin that was never run against the target before.

Below a simple example of the bug:
```bash
$ python2 ./owtf.py --nowebui -o OWTF-WVS-001 http://127.0.0.1 

 _____ _ _ _ _____ _____
|     | | | |_   _|   __|
|  |  | | | | | | |   __|
|_____|_____| |_| |__|

        @owtfp
    http://owtf.org
    
[-] Defaulting Plugin Group to 'web' based on list of plugins supplied
[*] OWTF Version: 1.0.1, Release: LionHeart (Beta) 
[-] Loading framework please wait..
[-] The IP address for 127.0.0.1 is: '127.0.0.1'
[*] 127.0.0.1:8008 <-- HTTP(S) Proxy to which requests can be directed
[-] __________ 1 - Target: http://127.0.0.1 -> Plugin: Arachni Unauthenticated (web/active) __________
# [omitted]
[-] __________ 2 - Target: http://127.0.0.1 -> Plugin: Arachni Unauthenticated (web/external) __________
# [omitted]
[-] All jobs have been done. Exiting.
[-] Stopping inbound proxy processes and cleaning up, Please wait!

$ python2 ./owtf.py --nowebui -o OWTF-WVS-006 http://127.0.0.1

 _____ _ _ _ _____ _____
|     | | | |_   _|   __|
|  |  | | | | | | |   __|
|_____|_____| |_| |__|

        @owtfp
    http://owtf.org
    
[-] Defaulting Plugin Group to 'web' based on list of plugins supplied
[*] OWTF Version: 1.0.1, Release: LionHeart (Beta) 
[-] Loading framework please wait..
[*] http://127.0.0.1 already exists in DB
[*] 127.0.0.1:8008 <-- HTTP(S) Proxy to which requests can be directed
[-] All jobs have been done. Exiting.
[-] Stopping inbound proxy processes and cleaning up, Please wait!
```

The bug seems to come from ``framework/config/config.py`` where ``Config.LoadTargets`` returns an empty if the target already exists in the DB (https://github.com/owtf/owtf/blob/develop/framework/config/config.py#L193).
[...]

PR #391, opened 2015-02-01
Reuse of OWTF against the samet target and plugin with -f option works

Sentient07:


framework/config/config.py
```diff
@@ -202,10 +202,15 @@ def LoadProfiles(self, profiles):
     def LoadTargets(self, options):
         scope = self.prepare_url_scope(options['Scope'], options['PluginGroup'])
         added_targets = []
+
         for target in scope:
+            added_targets.append(target)
+            force_option = False
+            if '-f' in options['argv']:
+                cprint("Test")
```
tunnelshade: You forgot a print statement?

Sentient07: Ouch, had that for testing :P Will remove it and send another one. Sorry !

framework/db/target_manager.py
```diff
@@ -163,11 +175,9 @@ def AddTarget(self, TargetURL, session_id=None):
             target_id = config_obj.id
             self.CreateMissingDBsForURL(TargetURL)
             self.SetTarget(target_id)
+
         else:
-            session_obj = self.db.session.query(
```
DePierre: By removing these lines above in a ``if`` statement introduces a bug. If the target URL is already know but the user does not use ``-f``, then OWTF falls in the ``else`` statement where ``target_obj`` does not exist, hence an error ``UnboundLocalError``.

DePierre:
@Sentient07 Could you take a look at my comment and fix it? Also could you revert your PEP8 commit because first, it is not the PEP8 OWTF tries to follow, second because it makes the review harder.

DePierre:
@Sentient07 Are you still working on this? I see no fix for the comment I have made on your commit.

delta24:
@DePierre Can you get me up to speed on this? I reverted the PEP8 commit locally.

DePierre:
@delta24 I'll check that ASAP to remember what was going on and get back to you

DePierre:
@delta24 You'll find the corresponding issues' description at #379 and #378. Let me know if you need more details but IMO the two issues should be enough to get you up to speed.

delta24:
Closing this PR for now.

PR #394, opened 2015-02-12
Updated python libraries  to latest versions 

Ahiknsr:


tunnelshade:
Solving this issue might have required some effort. Commits are cheap but that doesnt mean they are "Indian" cheap. :P . For such small change, please use single commit. Only versions had to be changed, even if some mistakes happened use --amend before pushing

PR #395, opened 2015-02-12
'aux' is no more considered target

Sentient07:


DePierre:
@Sentient07 This is not the way to go. Better leave the issue opened and wait for a proper fix.

PR #410, opened 2015-04-12
Added sslv2 support for kali and o-saft plugin

Ahiknsr:
Added o-saft as semi-passive plugin   and script to patch sslv2 of kali


DePierre:
Hi @Ahiknsr and thank you for this PR. I have two remarks before reviewing it though:

1. You should split this PR in two different ones: one for the sslv2 support and another for o-saft plugin.
2. Why isn't that SSLv2 patch submitted upstream? <http://blog.opensecurityresearch.com/2013/05/fixing-sslv2-support-in-kali-linux.html> that is quoted in the SSLv2 patch is from 2013... Why isn't it already merged into Kali upstream?

Cheers.

Ahiknsr:
@depierre , sslv2 support  was fixed in kali 1.1.0 [ https://bugs.kali.org/view.php?id=146 ] which was released in feb 9 ,2015 . i don't think all the users might have their systems upto date . so i have added this patch as a workaround

DePierre:
@Ahiknsr if it was merge upstream, OWTF should not have its own patch just in case imho.

PR #411, opened 2015-04-12
 Added O-saft tool to Testing for ssl-tls plugin in active category

Ahiknsr:


DePierre:
@Ahiknsr Thanks for the commit but doesn't OWTF need a plugin python file as well?

Because as-is, your commit only add an entry in the binaries known by OWTF but the user cannot use it since not python plugin exist for O-Saft.

PR #415, opened 2015-04-21
changed url for exploidb search command in passive plugin to reflect  latest changes in website .

Ahiknsr:


tunnelshade:
There is a small observation I made with the new exploit db. Try opening the following two links
+ https://www.exploit-db.com/search/?action=search&text=IIS
+ https://www.exploit-db.com/search/?action=search&description=IIS

So, instead of using `Free Text Search`, it is better to go with the `Title` search which is taken from get parameter `description`.

7a:
I'd suggest "text" because sometimes, the exact version affected is only shown on the advisory or top of the exploit, especially in cases when the affected component is a library or something like "also affects tomcat 1.2.3" -which WON'T be in the title, but is still relevant! :)-
So, imho, please back to:
https://www.exploit-db.com/search/?action=search&text=phpbb

tunnelshade:
Inner text search is not working for IIS alone :P.

DePierre:
The example provided by @tunnelshade convinced me that this is the way to go. Thank you @Ahiknsr for that patch :) Merged!

Issue #414, opened 2015-04-15
ValueError when OWTF is run without postgresql properly configured


DePierre:
When running OWTF with a DB not properly configured, a ``ValueError`` is thrown.

```pytb
root@kali32:~/owtf# python2 owtf.py 

 _____ _ _ _ _____ _____
|     | | | |_   _|   __|
|  |  | | | | | | |   __|
|_____|_____| |_| |__|

        @owtfp
    http://owtf.org
    
Traceback (most recent call last):
  File "owtf.py", line 300, in <module>
    main(sys.argv)
  File "owtf.py", line 263, in main
    ComponentInitialiser.initialisation_phase_1(root_dir, owtf_pid)
  File "/root/owtf/framework/dependency_management/component_initialiser.py", line 55, in initialisation_phase_1
    DB()
  File "/root/owtf/framework/db/db.py", line 68, in __init__
    self.create_session()
  File "/root/owtf/framework/db/db.py", line 97, in create_session
    self.Session = self.CreateScopedSession()
  File "/root/owtf/framework/db/db.py", line 146, in CreateScopedSession
    self.engine = self.CreateEngine(models.Base)
  File "/root/owtf/framework/db/db.py", line 133, in CreateEngine
    poolclass=NullPool)  # TODO: Fix for forking
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/__init__.py", line 346, in create_engine
    return strategy.create(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/strategies.py", line 49, in create
    u = url.make_url(name_or_url)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/url.py", line 164, in make_url
    return _parse_rfc1738_args(name_or_url)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/url.py", line 210, in _parse_rfc1738_args
    return URL(name, **components)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/engine/url.py", line 60, in __init__
    self.port = int(port)
ValueError: invalid literal for int() with base 10: ''
```

OWTF should instead indicate that the DB is not configured.

PR #451, opened 2015-05-03
Fixed for Issue #414

DoomTaper:


DoomTaper:
I have checked that value error is raised only beacuse of inappropriate "databse_port", so, we can also print out that this is due to inappropriate port in db config file to be more specific.

DePierre:
@DoomTaper Thanks for the fix, I have adapted your commit for the develop branch as well as added a print to show the current error in the ``DATABASE_PORT`` value.

See https://github.com/owtf/owtf/commit/2f8504ff8643b2673a846ccac7aa4d1cdc411531

PR #462, opened 2015-05-19
New install script to patch pycurl error on vanilla kali

g4l4drim:
missing curl-config command error message for pip install pycurl, due to package version depdendancies issues in kali-linux with libcurl4-openssl-dev

fixing by installing python-pycurl kali package and resume installing other required packages


delta24:
Closing as relevant now IMO

Issue #472, opened 2015-06-19
Timestamps not present in transaction log


tunnelshade:
It is a must to have timestamps in the transaction log, as it is sometimes necessary to know what occured first and what occured later

tunnelshade:
Fixed in 6cc46f099b68f8997daf328e691b37869e6302d6

PR #475, opened 2015-06-25
Fixes Issue #472

arunk-s:
Adding support for timestamps in Transaction Logs and `datetime` object serialization.

### DIFF
--- a/framework/db/models.py
+++ b/framework/db/models.py
@@ -90,6 +90,7 @@ class Transaction(Base):
     data = Column(String, nullable=True)  # Post DATA
     time = Column(Float(precision=10))
     time_human = Column(String)
+    local_timestamp = Column(DateTime)
     raw_request = Column(Text)
     response_status = Column(String)
     response_headers = Column(Text)
--- a/framework/db/transaction_manager.py
+++ b/framework/db/transaction_manager.py
@@ -198,6 +198,7 @@ class TransactionManager(BaseComponent, TransactionInterface):
                 trans.response_status,
                 str(trans.time),
                 trans.time_human,
+                trans.local_timestamp,
                 trans.data,
                 trans.raw_request,
                 trans.response_headers,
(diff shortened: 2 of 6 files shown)


tunnelshade:
This will not work when the transactions are logged by transaction logger. The actual time and date has to be generated when saving the transaction cache i.e in the proxy. This will not work

tunnelshade:
This will require some background work as well, as our proxy is multi instance there will be lots of requests at the same time, so the accuracy of the time has to be higher while not slowing down the proxy.

7a:
Why not let the database do the work?
sysdate, current_timestamp, etc. Every decent DB has a function for local
timestamp.

Technically, this would be time of logging, not exact time of the
transaction. However, a transaction is a request-response pair, and
therefore there can be a delay of a few seconds, sometimes minutes (on
heavy screens), from the time the request is sent until the time the
response is received.

With this in mind: Is the DB timestamp, at the transaction logging stage a
delay we can live with, or is there a better way? Do we need more precision?

tunnelshade:
The proxy has to do the datetime save. Transaction logger takes request-response pairs from proxy cache and put them into our db. So, when we use our cache manager to cache our http requests we should imo save the time that we got the response and use it for ordering our transaction log.

From my profiling the delay is negligible. But few others should try profiling proxy before and after adding this functionality to our cache manager

DePierre:
@tunnelshade Any thought on that? You know that stuff way better than me and I can't properly evaluate the changes.

tunnelshade:
@DePierre Looks fine. @arunk-s You tested all the transaction search features? Are you sure you didn't break anything?

arunk-s:
@tunnelshade I did checked all the transaction search features, they all are working.

DePierre:
@arunk-s @tunnelshade I will test that this week-end if I have enough time. Otherwise on Wednesday. If anyone has more free time before, don't be shy.

tunnelshade:
Thanks for the PR @arunk-s. I imported your changes and corrected few of them and added few more @ 6cc46f099b68f8997daf328e691b37869e6302d6

PR #480, opened 2015-06-30
Remove user input for distro detection

alexandrasandulescu:
Add support in install/install.py for automatically distro detection

### DIFF
--- a/install/install.py
+++ b/install/install.py
@@ -28,6 +28,7 @@
 #
 import os
 import time
+import platform
 from datetime import datetime
 
 import ConfigParser
@@ -77,28 +78,40 @@ class Installer(object):
             self.install_in_directory(os.path.expanduser(cp.get(section, "directory")), cp.get(section, "command"))
 
     def install(self):
-        # User asked to select distro and distro related stuff is installed
+        # User asked to select distro (in case it cannot be automatically detected) and distro related stuff is installed
         cp = ConfigParser.ConfigParser({"RootDir":self.RootDir, "Pid":self.pid})
         cp.read(self.distros_cfg)
 
+        #Try get the distro automatically
+        distro, version, arch = platform.linux_distribution()
+        distro_num = 0
+        if "kali" in distro.lower():
+            distro_num = 1
+        elif "samurai" in distro.lower():
+            distro_num = 2
+
         # Loop until proper input is received
         while True:
+            if distro_num != 0:
+                print("[*] %s has been automatically detected... Continuing in auto-mode"%(distro))
+                break
             print("")
             for i in range(0, len(cp.sections())):
                 print("(%d) %s"%(i+1, cp.sections()[i]))
-            print("(0) %s"%("My distro is not listed :("))
+            print("(0) %s (%s)"%("My distro is not listed :(", distro))
             distro_num = raw_input("Select a number based on your distribution : ")
             try: # Cheking if valid input is received
                 distro_num = int(distro_num)
-                if distro_num != 0:
-                    self.run_command(cp.get(cp.sections()[int(distro_num)-1], "install"))
-                else:
-                    print("Skipping distro related installation :(")
                 break
             except ValueError:
                 print('')
                 print("Please enter a valid number")
                 continue
+
+        if distro_num != 0:
+            self.run_command(cp.get(cp.sections()[int(distro_num)-1], "install"))
+        else:
+            print("Skipping distro related installation :(")
 
         # First all distro independent stuff is installed
         self.install_restricted_from_cfg(self.restricted_cfg)


install/install.py
```diff
@@ -77,28 +78,25 @@ def install_restricted_from_cfg(self, config_file):
...
+            status, result = commands.getstatusoutput(command)
+        distro_num = 0
+        if "kali" in result.lower():
+            distro_num = 1
+        elif "samurai" in result.lower():
+            distro_num = 2
+
+        if distro_num != 0:
+            self.run_command(cp.get(cp.sections()[int(distro_num)-1], "install"))
+        else:
```
tunnelshade: If we cannot detect the distro, please prompt the user to select one. There are cases when someones takes a distribution and then add few new sources and customizations and rename it or someone using debian but has all the kali repos. So, if we are unable to detect known distros prompt the user to select one

alexandrasandulescu: I have modified the commit and now it includes the changes you mentioned.

install/install.py
```diff
@@ -77,12 +78,29 @@ def install_restricted_from_cfg(self, config_file):
             self.install_in_directory(os.path.expanduser(cp.get(section, "directory")), cp.get(section, "command"))
 
     def install(self):
-        # User asked to select distro and distro related stuff is installed
+        # User asked to select distro (in case it cannot be automatically detected) and distro related stuff is installed
         cp = ConfigParser.ConfigParser({"RootDir":self.RootDir, "Pid":self.pid})
         cp.read(self.distros_cfg)
 
+        #Try get the distro using lsb_release
+        status, result = commands.getstatusoutput("lsb_release -si")
```
tunnelshade: Commands module is removed in python3. It will cause portability issues. Use subprocess and we can merge this

DePierre:
Maybe we could do that in python? I know, that's silly...
```python
import platform
platform.linux_distribution()
```
[Python 2.X](https://docs.python.org/2.7/library/platform.html#platform.linux_distribution) or [Python 3.X](https://docs.python.org/3.4/library/platform.html#platform.linux_distribution)

tunnelshade:
Yup. Check [this](http://stackoverflow.com/questions/4553129/when-to-use-os-name-sys-platform-or-platform-system) for more info :P

alexandrasandulescu:
I tried to use sys.platform() and platform.system() but I don't receive the distro, but the os type (as it says in description - Linux, Windows, etc).
os.uname() seems ok but I had a lot of problem with the distro it provides(for example on centos releases). The distribution must be parsed from the output also.

I suggest using subprocess to avoid compatibility issues and assure that the output is correct
What do you think about it?

DePierre:
@alexandrasandulescu I think that if python provides such functions to get the distribution, and that such functions have been standard since python 2.x, we might want to use them instead of making our own soup.

I am opened to discussion.

alexandrasandulescu:
hi @DePierre ,
I replaced command with subprocess as @tunnelshade  mentioned and the methods used now (check_call, check_output) are available since python 2.7 [1] . Does it break compatibility?
I will switch to 'uname ' if it does.
I am not against using it. I tested in kali and it prints the correct distro. I will have to test in samuraiwtf too.

Thanks

[1]https://docs.python.org/2/library/subprocess.html

DePierre:
@alexandrasandulescu I have more faith in python's standard functions that gives you the distribution information than a home-made bash script.

Also, I don't like calling bash script from python source code. If we are writing python, let's write python.

Why would we prefer home-made bash script called from within python using ``subprocess`` manually than using python standard functions that do all the work already and have been tested since 10 years?

@tunnelshade What's your feeling about that?

alexandrasandulescu:
@DePierre  @tunnelshade 
I used platform.linux_distribution()
How does it look now?
Thanks

tunnelshade:
`platform.linux_distribution` reads from `/etc/lsb-release`. What about non debian based distros?

DePierre:
@tunnelshade That would not be true. According to their [source code](https://hg.python.org/cpython/file/3.5/Lib/platform.py#l309):
```python
"""
        The function first looks for a distribution release file in
        /etc and then reverts to _dist_try_harder() in case no
        suitable files are found.
"""
```
Notice the "a distribution file" meaning not only ``/etc/lsb-release``. And with ``_dist_try_harder()`` implementing [other tricks](https://hg.python.org/cpython/file/3.5/Lib/platform.py#l203) to find the distribution.

If you meant that you want to check for MacOS or Windows as well, better support them first, and then add the checks imho.

alexandrasandulescu:
According to [1] the supported distributions are 
'SuSE', 'debian', 'fedora', 'redhat', 'centos', 'mandrake', 'mandriva', 'rocks', 'slackware', 'yellowdog', gentoo', 'UnitedLinux', 'turbolinux'.

[1] https://hg.python.org/cpython/file/2.7/Lib/platform.py#l259

tunnelshade:
In arch linux, it only returns 'arch' which might be painful in future.

@DePierre I read that statement from stackoverflow. I should have checked the source.

Not everything is covered under this linux_distribution method. Any ideas? @DePierre

DePierre:
@tunnelshade my 2 cents: Python's standard functions will cover more cases than doing some home made tests.
As for the ``arch``, I don't see much of a big deal here. Even with only ``arch`` we can update the install script accordingly.

We can still rollback to ``platform.platform()`` later. We will have plenty of time before reaching the limits of ``platform.linux_distribution()``. One day we will have to add support for Windows and/or Mac.

I will test @alexandrasandulescu's patch.

DePierre:
@alexandrasandulescu Would it be possible to tell the user that OWTF automatically detected X platform and is going to install automatically.

For instance something like ``[*] Kali has been automatically detected... Continuing in auto-mode`` or something. Otherwise, your PR is good to me :)

alexandrasandulescu:
@DePierre done .

DePierre:
@alexandrasandulescu Looks good to me and I am ready to merge it as-is. Nice feature you implement here :)

Any thought @tunnelshade?

tunnelshade:
Ok :+1:

DePierre:
@alexandrasandulescu Thank you for your contribution ;)

alexandrasandulescu:
thank you for your help and guidance! @all

PR #484, opened 2015-07-13
Fixed minor typos

MarkVLK:
Just saw these typos while I was installing OWTF and figured I'd correct them because I have OCD


DePierre:
@MarkVLK Thanks, appreciate that ;)

PR #487, opened 2015-07-17
Add db config file path parameter for scripts/db_run.sh

alexandrasandulescu:


scripts/db_run.sh
```diff
@@ -56,7 +75,11 @@ SCRIPTS_DIR=$(dirname "$FILE_PATH")
 RootDir=${2:-$(dirname "$SCRIPTS_DIR")}
 
 config_file="$RootDir/framework/config/framework_config.cfg"
-db_config_file="$(get_config_value DATABASE_SETTINGS_FILE $config_file)"
+db_config_file="$(get_db_config_file $config_file $1)"
```
tunnelshade: The usage of RootDir is for getting default path of the framework config. I think it is better to take an optional parameter of the **framework_config.cfg**. The reason being that there has to be right db_config_file path in framework config or else OWTF will not work. So we only need one optional cmd parameter for this script (i.e framework_config.cfg). Please let me know if I am not clear

alexandrasandulescu:
@tunnelshade I removed the second command line parameter :)

PR #488, opened 2015-07-20
Distro independent PostgreSQL daemon start

alexandrasandulescu:


scripts/db_run.sh
```diff
@@ -45,6 +45,30 @@ get_postgres_server_port() {
...
+
+    INIT_COMMAND="initdb --locale en_US.UTF-8 -E UTF8 -D '$PGDATA'"
+    su $PGUSER -c "$INIT_COMMAND"
+
+    echo "[+] Starting PostgreSQL ..." 1>&2
+    START_COMMAND="postgres -D '$PGDATA' &"
+    su $PGUSER -c "$START_COMMAND" >>$PGLOG 2>$PGLOG
+
+    #Check the logfile for errors.
+    status=$(cat $PGLOG | grep "fail" | wc -l)
```
tunnelshade: The output gets appended to log right? So any failed old runs might trigger this condition?

alexandrasandulescu:
If systemctl or service cannot start the daemon, then the data directory should be empty or missing (PGDATA). So theoretically the serverlog file will contain the logs for the first start of the server. If you think it is safer too use another log file, I will add the change

tunnelshade:
I am only talking about the log file. Any old runs that failed might have added the word fail to the log file right?

tunnelshade:
Hey @alexandrasandulescu Did you try this in Kali? I don't think there is an **initdb** in Kali

scripts/db_run.sh
alexandrasandulescu: Not really. PGLOG is recreated before initdb (otherwise initdb fails)

alexandrasandulescu:
@tunnelshade I tried this in arch (in Kali, this path won't be executed because service or systemctl will do their job). Even in arch it could not be executed if the daemon exists in systemctl . Basically, the problem is that #pacmam -S postgresql  doesn t take take of the postgresql daemon setup (add the service in systemctl). In Debian based distros, aptitude takes care of everything and the daemon can be started without initdb. 

Also, initdb is a postgres commmand , it can be executed on any distro.

tunnelshade:
initdb is not present in Kali @alexandrasandulescu

alexandrasandulescu:
@tunnelshade , tested, works on both kali and arch

scripts/start_postgres.sh
```diff
@@ -0,0 +1,51 @@
...
+PGUSER=postgres
+PGVAR="/var/lib/postgres"
+if [ ! -d "$PGVAR" ]; then
+  PGVAR="/var/lib/postgresql"
+fi
+
+PGDATA="$PGVAR/data"
+PGLOG="$PGDATA/serverlog"
+
+if [ -d "$PGDATA" ]; then
```
tunnelshade: User might be having some data XD

PR #489, opened 2015-07-21
Extend dependency check for modules installed via package manager

alexandrasandulescu:


framework/dependency_check.py
```diff
@@ -66,6 +73,11 @@ def verify_dependencies(root_dir):
         if library_name not in installed_libraries.keys():
             missing_libraries.append(library_name)
 
+    # Check if the missing modules are installed via system package manager
+    for lib in missing_libraries:
```
tunnelshade: Modifying the same list while iterating it is not a good idea!

alexandrasandulescu:
@tunnelshade fixed :)

PR #486, opened 2015-07-17
Check access rights for OUTPUT_PATH

alexandrasandulescu:
I hardcoded ~/.owtf. Where should this default path be added? default.cfg is not a good idea because users can modify it and we want "~/.owtf" to be the failure solution for bad config.

PR #490, opened 2015-07-24
OWTF AUR package

alexandrasandulescu:
what do you suggest regarding sha1sums? should exist any checks? also, I mention that this change can be merged only after merging #486(mandatory) and #488  (optional).


tunnelshade:
Ok. I just have some doubts with the postgres fix. I will check it and merge all these

tunnelshade:
For sha1sums we should be able to automate the generation of sha1sums or else it is too much of a maintanance right?

PR #493, opened 2015-07-31
Fix binding address for FILE_SERVER_PORT

alexandrasandulescu:


framework/interface/server.py
```diff
@@ -62,7 +62,7 @@ def start(self):
                 int(config.FrameworkConfigGet(
                     "FILE_SERVER_PORT")),
                 address=config.FrameworkConfigGet(
-                    "SERVER_ADDR")
+                    "UI_SERVER_ADDR")
```
tunnelshade: I don't see any **UI_SERVER_ADDR** [here](https://github.com/owtf/owtf/blob/develop/framework/config/framework_config.cfg#L18)

delta24: The fix is to replace the 0.0.0.0 in the `interface/server.py` to 127.0.0.1

alexandrasandulescu:
tested

tunnelshade:
Thanks @alexandrasandulescu

PR #497, opened 2015-08-08
Develop

DoomTaper:
just edited "unable to add" instead of "unable add"


tunnelshade:
This not a clean PR, for such a small change create only one commit. Here the merge commit is extra. The commit message has to explain the fix properly. Check [this](https://github.com/owtf/owtf/commit/451c610f35ceb0380d1365c1e7e4490e2787cbd4)

Issue #399, opened 2015-02-24
replace msfcli with msfconsole -x or -r


anantshri:
We need to replace msfcli with its newer counterparts.

https://github.com/rapid7/metasploit-framework/issues/3802
<pre>
[!] ************************************************************************
[!] *               The utility msfcli is deprecated!                      *
[!] *            It will be removed on or about 2015-06-18                 *
[!] *              Please use msfconsole -r or -x instead                  *
[!] *  Details: https://github.com/rapid7/metasploit-framework/pull/3802   *
[!] ************************************************************************
</pre>

7a:
+1

qgrosperrin:
I'm running into some issues using the `msfconsole -x` alternative to `msfcli` and was hoping to get some directions here on how to use it properly. To me, it appears:
- 1. slower than msfcli  
- 2. too verbose  
- 3. the command line doesn't look very clean

Below are some details about these comments.

1. 
I have noted that the `msfconsole` alternative to `msfcli` is at least two times slower on my machine. I'm guessing this is because `msfconsole` tries to load everything, not taking into account it runs as a command line utility (as opposed to interactive mode). Is there a way to use `msfconsole -x/-r` and not have it load everything ?

Sample exec times from my machine
```
time msfcli auxiliary/scanner/finger/finger_users USERS_FILE=/usr/share/wordlis
ts/metasploit/unix_users.txt RHOSTS=192.168.231.145 E
[...]
real    0m4.412s
user    0m3.288s
sys     0m1.080s
```

```
time msfconsole -x "use auxiliary/scanner/finger/finger_users; set USERS_FILE /
usr/share/wordlists/metasploit/unix_users.txt; set RHOSTS 192.168.231.145; exploit; exit"
[...]
real    0m11.185s
user    0m5.860s
sys     0m5.244s
```

2. `msfconsole` prints the metasploit framework welcome message and logo which is a lot more verbose than `msfcli` but this does not provide additional technical details.. Is there a way to have `msfconsole` not print all that information and instead just output the essential information that `msfcli` was exclusively returning ?

You might think this is not really important but I usually have `msfcli` run inside a wrapper function (BASH, Python or else) so having unnecessary verbose output can sometimes be a problem.

3. Is there any ways to have the command below executed in one go (not using ";" separated stacked commands), similar to msfcli ? 
[...]

PR #502, opened 2015-08-14
Replace msfcli with msfconsole

alexandrasandulescu:
Issue #399 partial fix for Ftp, Smtp, Vnc, Emc, MsSql, MsRpc, HttpRpc,
Smb, Snmp Probe Methods


sealmindset:
I just comb through all the branches from adventure through ui looking to see if any changes were made to convert msfcli to msconsole dash x. So far all of the branches are still using the now defunct metasploit cmd. My workaround is to do a replacement:

grep -rl msfcli '/path/to/owtf' | xargs sed -i 's/msfcli /msfconsole -x /g'

Which affected these files:

/owtf/ui/owtf//profiles/general/default.cfg
/owtf/ui/owtf//profiles/general/default_backtrack.cfg
/owtf/ui/owtf//profiles/resources/default.cfg
/owtf/ui/owtf//profiles/resources/default_backtrack.cfg
/owtf/ui/owtf//scripts/import/metasploit/import_exploits.sh

It seems to be an easy fix.

PR #504, opened 2015-08-18
Add available space checking mechanism

alexandrasandulescu:
* add wget wrapper to check if the download can proceed
* install tools only after enough space is available

### DIFF
--- a/install/distro-independent.cfg
+++ b/install/distro-independent.cfg
@@ -10,7 +10,7 @@ command = command -v pip2 >/dev/null || { wget https://bootstrap.pypa.io/get-pip
 
 [Httprint]
 directory = %(RootDir)s/tools/restricted/httprint
-command = wget "http://www.net-square.com/zip folders/httprint_linux_301.zip"; unzip *.zip; rm -f *.zip
+command = wget http://www.net-square.com/_assets/httprint_linux_301.zip; unzip *.zip; rm -f *.zip;
 			cp -rf %(RootDir)s/tools/httprint-signatures.txt %(RootDir)s/tools/restricted/httprint/httprint_301/linux
 
 [BIG-IP Decoder]
--- a/install/install.py
+++ b/install/install.py
@@ -32,6 +32,7 @@ import time
 import platform
 import argparse
 from datetime import datetime
+from space_checker_utils import wget_wrapper
 
 import ConfigParser
 
@@ -61,6 +62,8 @@ class Installer(object):
     def run_command(self, command):
         print("[*] Running following command")
         print("%s"%(command))
+        if not wget_wrapper(command):
+            return
         os.system(command)
 
     def install_in_directory(self, directory, command):
--- a/install/space_checker_utils.py
+++ b/install/space_checker_utils.py
@@ -1,0 +1,94 @@
+from __future__ import division
+import re
+import os
+import subprocess
+import urllib2
+
+
+def byte_to_human(size):
+    """Transform the size in bytes into the appropriate unit of measurement.
+
+    :param int size: Size in bytes.
+
+    :return: Actual size in B/KB/MB/GB.
+    :rtype: str
+    """
+    try:
+        size = int(size)
+    except ValueError:
+        return (0, "B")
+    um = "B"
+    if size < pow(2,20):
+        size = size / pow(2,10)
+        um = "KB"
+    elif size < pow(2,30):
+        size = size / pow(2, 20)
+        um = "MB"
+    else:
+        size = size / pow(2, 30)
+        um = "GB"
+
+    return (size, um)
+
+
+def get_wget_download_size(command):
+    """Get the whole download size in bytes.
+
+    :param str command: Wget command.
+
+    :return: The actual download size in bytes.
+    :rtype: int
+    """
+    # Wget supports HTTP, HTTPS and FTP protocols.
+    if not command.startswith('wget '):
+        return 0
+
+    # Check the url(s) is/are valid.
+    urls = re.findall(r'((?:https|http|ftp)://[^\s]+)', command)
+
+    if not urls:
+        return 0
+
+    size = 0
+    for url in urls:
+        try:
+            response = urllib2.urlopen(url)
+            size += int(response.info().getheader('Content-Length'))
+        except (urllib2.HTTPError, urllib2.URLError, ValueError):
+            pass
+    return size
+
+
+def get_fs_free_space():
+    """Get the available space of the current filesystem.
+
+    :return: The available size in KB.
+    :rtype: int
+    """
+    try:
+        stat = os.statvfs('.')
+        return (stat.f_bavail * stat.f_frsize) / 1024
+    except  OSError:
+        print("[!] Failed to get the filesystem disk space usage")
+        return 0
+
+
+def wget_wrapper(command):
+    """Checks if the current filesystem has enough available space in order to proceed downloading data using wget.
+
+    :param str command: The installation command.
+
+    :return: The installation can proceed.
+    :rtype: bool
+    """
+    commands = re.findall('(wget.*?);', command)
+    size = sum(get_wget_download_size(wget_command) for wget_command in commands)
+
+    while size / 1024 > get_fs_free_space():
+        print("[!] Not enough space to proceed with the download.")
+        human_size, um = byte_to_human(size)
+        answer = raw_input("[!] Please free %s %s and proceed, or type \'n\' to skip the download. [Y/n] "
+                %('{0:.3g}'.format(human_size), um))
+        if answer == 'n':
+            return False
+    return True


install/install.py
```diff
@@ -61,6 +62,9 @@ def create_directory(self, directory):
     def run_command(self, command):
         print("[*] Running following command")
         print("%s"%(command))
+        if is_wget(command):
```
DePierre: ``wget_wrapper`` begins with:
```python
if not is_wget(command):
    return True
```

So it does not look necessary to test it here.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
+from __future__ import division
+import re
+import os
+import subprocess
+
```
DePierre: Missing blank line on level-0 indent.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+def get_wget_download_size(command):
+    """Get the whole download size.
```
DePierre: In what unit?

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    :return: The actual download size.
+    :rtype: int
+    """
+    try:
+        _, args = command.split(' ', 1)
+    except IndexError:
+        return 0
+
+    download_size_command = "wget --spider --server-response " + args
```
DePierre: Could use '%' or ``format()`` instead of concat.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    proc = subprocess.Popen(download_size_command.split(), stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT)
```
DePierre:
```python
proc = subprocess.Popen(
    download_size_command.split(),
    stdout=subprocess.PIPE,
    stderr=subprocess.STDOUT)
```

This is actually a personal preference. It is easier to only add one more indentation level when organizing the parameters of a long function call. It is also easier to read than read 2 parameters, then read 1 parameter.

Not mandatory, just a feeling from previous experience with pep8

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    for line in proc.stdout:
```
DePierre: It would be nicer to use a regex in this case, where you would at once:
1. Find the correct HTTP header
2. Check that the value of the HTTP header is an integer
3. Retrieve that value (using regex groups)

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        value = size / pow(2, 20)
+        um = "MB"
+    else:
+        value = size / pow(2, 30)
+        um = "GB"
+
+    return str('{0:.3g}'.format(value)) + " " + um
+
+def is_wget(command):
+    return "wget" in command
```
DePierre: This looks way too lose to be useful. What if the command is ``iwgetid`` for instance? Or what if the command is ``rm pr0n/midwget/*``? (I didn't anything else than midget that would look like wget sorry).

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+def get_fs_free_space():
+    """Get the available space of the current filesystem.
+
+    :return: The available size in KB.
+    :rtype: int
+    """
+
```
DePierre: Unecessary empty line

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        except (IndexError, ValueError):
+            continue
+        if key == "Content-Length":
+            try:
+                size = int(value)
+                return size
+            except ValueError:
+                break
+    return 0
+
```
DePierre: Missing blank line on level-0 indent.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        um = "KB"
+    elif size < pow(2,30):
+        value = size / pow(2, 20)
+        um = "MB"
+    else:
+        value = size / pow(2, 30)
+        um = "GB"
+
+    return str('{0:.3g}'.format(value)) + " " + um
+
```
DePierre: Missing blank line on level-0 indent.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+        value = size / pow(2,10)
+        um = "KB"
+    elif size < pow(2,30):
+        value = size / pow(2, 20)
+        um = "MB"
+    else:
+        value = size / pow(2, 30)
+        um = "GB"
+
+    return str('{0:.3g}'.format(value)) + " " + um
```
DePierre: Why using ``+`` when you are already using format? You could have something like: ``return '{0:.3g} {1}'.format(value, unit)``.

Also why the ``str()``? Since format is a string method, it necessarily returns a string.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+def is_wget(command):
+    return "wget" in command
+
```
DePierre: Missing blank line on level-0 indent.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+def get_fs_free_space():
+    """Get the available space of the current filesystem.
+
+    :return: The available size in KB.
+    :rtype: int
+    """
```
DePierre: What about the exception ``os.statvfs`` can raise? Such as ``OSError``? Whether you document them or you catch them. In your case, you should catch them IMO.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+
+    stat = os.statvfs('.')
+    return (stat.f_bavail * stat.f_frsize) / 1024
+
```
DePierre: Missing blank line on level-0 indent.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+def wget_wrapper(command):
+    """Checks if the current filesystem has enough available space
```
DePierre: Remember, 120chars max. In your vim, please update ``set textwidth=119``, that might help you with that.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    in order to proceed downloading data using wget.
+
+    :param str command: The installation command.
+
+    :return: The installation can proceed.
+    :rtype: bool
+    """
+
```
DePierre: Unecessary empty line

install/space_checker_utils.py
DePierre: Unecessary empty line

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+def size_to_str(size):
+    """Transform size in bytes into the appropriate unit of measurement.
+
+    :param int size: Size in bytes.
+
+    :return: Actual size in B/KB/MB/GB.
+    :rtype: str
+    """
+
```
DePierre: Unecessary empty line

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    try:
+        size = int(size)
```
DePierre: The only place where you call ``size_to_str`` is in ``wget_wrapper`` where ``size`` has been asserted to be an integer, so the ``try/catch`` is unecessary here.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+            continue
+        if key == "Content-Length":
+            try:
+                size = int(value)
+                return size
+            except ValueError:
+                break
+    return 0
+
+def size_to_str(size):
```
DePierre: IMO this function should take a size in bytes, and return to highest unit possible. It should try to cast it as a string.
Then when you call it in ``wget_wrapper`` you use ``%d`` or ``format()`` to convert it to a string. (You would need to rename the function to something like ``byte_to_human()``).

This would be nicer that way IMO.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    :rtype: str
+    """
+
+    try:
+        size = int(size)
+    except ValueError:
+        return ""
+    um = "B"
+    if size < pow(2,20):
+        value = size / pow(2,10)
```
DePierre: What not keep the variable name ``size``? It makes more sense than ``value`` and is more explicit.

install/space_checker_utils.py
```diff
@@ -0,0 +1,99 @@
...
+    :return: The installation can proceed.
+    :rtype: bool
+    """
+
+    if not is_wget(command):
+        return True
+    commands = re.findall(r'wget(.*?);', command)
+    size = 0
+    for command in commands:
+        size += get_wget_download_size(command)
```
DePierre: This could be more pythonesc with something like:
```python
size = sum(get_wget_download_size(command) for command in commands)
```

install/space_checker_utils.py
alexandrasandulescu: @DePierre , if the function will be used in the future by someone else, in another file or whatever, then the check makes sense. This function is not a very specific one.

alexandrasandulescu:
@DePierre , fixed

alexandrasandulescu:
@DePierre , fixed. I added some url checks in is_wget

install/distro-independent.cfg
```diff
@@ -10,7 +10,7 @@ command = command -v pip2 >/dev/null || { wget https://bootstrap.pypa.io/get-pip
 
 [Httprint]
 directory = %(RootDir)s/tools/restricted/httprint
-command = wget "http://www.net-square.com/zip folders/httprint_linux_301.zip"; unzip *.zip; rm -f *.zip
+command = wget http://www.net-square.com/_assets/httprint_linux_301.zip; unzip *.zip; rm -f *.zip
```
tunnelshade: Shouldn't there be a semicolon at the end as there is another command at the end?

install/space_checker_utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+    # Check the url(s) is/are valid.
+    urls = re.findall(r'((?:https|http|ftp)://[^\s]+)', command)
+
+    if not urls:
+        return False
+
+    for url in urls:
+        try:
+            urllib2.urlopen(url)
```
tunnelshade: You are doing a urllib2.urlopen call here, so you can just use ``response.info().getheader('Content-Length')``. Why the pain of using wget?

PR #508, opened 2015-08-21
Add aptitude wrapper support for kali script

alexandrasandulescu:
* used in install/kali/install.sh
* check available disk space before installing packages using apt-get

### DIFF
--- a/install/aptitude-wrapper.sh
+++ b/install/aptitude-wrapper.sh
@@ -1,0 +1,52 @@
+#!/usr/bin/env sh
+# This script is a wrapper over apt-get that installs
+# the packages only after checking the disk available space
+#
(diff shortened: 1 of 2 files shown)


install/aptitude-wrapper.sh
```diff
@@ -0,0 +1,43 @@
...
+    read yn
+    if [ "$yn" = 'n' ]; then
+        exit 0
+    fi
+done
+
+# If all the checks passed, install the required packages.
+for package in "$@"
+do
+    apt-get install $package
```
tunnelshade: You didn't check for root privileges, will this work?

alexandrasandulescu: @tunnelshade , fixed

Issue #525, opened 2015-12-06
Unable to open directory from browser


delta24:
Reproduce
---
When you pass an OWTF target containing especial URL characters such as "?" on the file browser (i.e. "Browse Files" button) on any plugin, this fails to load because these characters are not urlencoded, hence the browser interprets them and OWTF fails to open the relevant directory.

Fix
---
Calling the relevant url encoding function will fix this problem, for example using `encodeURIComponent` from JavaScript.

 or

calling `escape` from Tornado template methods

delta24:
The previous fix included using `quote_plus` from the urllib library. But it encoded '/' also, which resulted in bad links in the directory listing.
@DePierre thoughts on this?

DoomTaper:
@delta24 have you tried using `safe` parameter of quote_plus like this `quote_plus(item, safe='/')` ?

delta24:
No, that looks doable. Let me try :)
Thanks @DoomTaper!

DoomTaper:
@delta24 Is this issue resolved?

delta24:
nope, go ahead. :)

PR #538, opened 2016-01-20
fixes bad url file handling issue #525

DoomTaper:


framework/interface/templates/plugin_report.html
```diff
@@ -94,7 +95,7 @@ <h4 id="header_badge_{{ test_code }}"></h4>
                                         <td> {{ poutput['start_time'] }} <br />{{ poutput['end_time'] }} </td>
                                         <td> {{ poutput['status'] }} </td>
                                         {% if poutput['output_path'] %}
-                                            <td><a href="/output_files/{{ poutput['output_path'] }}/" target="_blank" class="btn btn-primary">Browse</a></td>
+                                            <td><a href="/output_files//{{ quote_plus(poutput['output_path'], safe='/') }}/" target="_blank" class="btn btn-primary">Browse</a></td>
```
delta24: Why the extra `/` ?

framework/interface/ui_handlers.py
```diff
@@ -440,11 +441,12 @@ class FileRedirectHandler(custom_handlers.UIRequestHandler):
     SUPPORTED_METHODS = ('GET')
 
     def get(self, file_url):
+	file_url = file_url[1:]
```
delta24: Intendation.

framework/interface/custom_handlers.py
```diff
@@ -34,7 +35,7 @@ def get(self, path, include_body=True):
         This is an edited method of original class so that we can show
         directory listing and set correct Content-Type
         """
-        path = self.parse_url_path(path)
+	path = self.parse_url_path(path)
```
delta24: Indentation.

framework/interface/templates/plugin_report.html
DoomTaper: By intercepting the request in burp suite I observed that if I dont put /(or any other character) then it is not making the correct GET request but I am not able reproduce it now :confused: .You check that @delta24  and if everything is fine make the appropriate changes.

framework/interface/ui_handlers.py
DoomTaper: It doesn't show any indentation error in my kali and also doesn't show any false indentation in my develop branch files. I don't why it is showing in the commit.

delta24: :P
Check once more in any other editor.

DoomTaper: I have already checked them on Gvim, Vim, gedit and leafpad, they were not showing any indentation problem but just now I checked it on sublime text 3 and it shows the indentation problem :P

delta24: It probably means there was a mixed indentation (tabs and spaces). Always use spaces. :)

framework/interface/templates/plugin_report.html
delta24: Remove `file_url = file_url[1:]` and you will not need it.

DoomTaper: Ya! it was due to that , my bad!! So we doesn't require it. thanks @delta24

framework/interface/ui_handlers.py
DoomTaper: thanks @delta24

delta24:
Wrong approach. I am committing the correct fix in a few minutes.

PR #539, opened 2016-01-25
Allow the install to proceed is user is root

dougmorato:
A change recently introduced check if the user has "sudo" permissions.

The check_sudo() function should also allow the installation to proceed if the user is root.

This Pull Requests checks  if the user has UID = 0 (equivalent to root) and proceed if it does have so, otherwise, print's the warning and exits cleanly.

### DIFF
--- a/install/install.py
+++ b/install/install.py
@@ -56,7 +56,7 @@ class Installer(object):
     def check_sudo():
         """Checks if the user has sudo access"""
         sudo = os.system("sudo -v")
-        if sudo == "0":
+        if sudo == 0:
             return
         else:
             print("[WARNING] Your user does not have sudo priveleges. Some OWTF components require sudo permissions to install")


DoomTaper:
@dougmorato , `os.system("sudo -v")` will return 0 if the user is root. So everything seems fine to me.

dougmorato:
Running on Kali 2.0, "sudo -v" didn't return anything, even though user was root. Thus, the submission of a PR.

Attached is screenshot showing empty response:

image

DoomTaper:
@dougmorato 
screenshot 47

anantshri:
I suppose the command should be os.getuid() or os.geteuid() and i can confirm that this solve the issue.

dougmorato:
@anantshri You are correct, I tested it on Kali but performed the change in my other machine which has my github credentials... Mistyped when committing the code, so updated the change and re-uploaded.

dougmorato:
@DoomTaper I believe then what needs to be changed is the Double-Quotes (") around the == "0" because the return from os.system("sudo -v") is an int

What do you prefer? Should I change my PR to remove the double-quotes then instead of using the elif geteuid  ?

delta24:
`os.getuid()` is not portable across different OSes hence the `sudo` check. Nice catch :)

Issue #448, opened 2015-05-02
Support OWTF on Windows


anantshri:
Consider this a meta issue opened to start gradual support of OWTF in windows.

starting we have following two major problems

https://github.com/owtf/owtf/tree/master/dictionaries/fuzzdb/fuzzdb-1.09/docs/misc/unicode-security contains a file and folder which contains ":" which makes it not creatable in windows.

simmilarly we use "aux" as a type of plugin which is a protected name in windows.

I wanted to work on this and hence started with my own fork of owtf however seems like i am not getting time and this project keep getting more and more commits in it. so can i ask someone to please have a look at these two issues, Once that is done we would be able to atleast clone owtf in windows and then we can start with testing for what is working and what not and the slowly look at supporting windows also.

7a:
+1

anantshri:
Just asking do we need this folder https://github.com/owtf/owtf/tree/master/dictionaries/fuzzdb/fuzzdb-1.09/docs/misc/unicode-security can we remove this folder all together from the core. That should help out a lot.

This will leave with issues like net and aux not creating coz they are reserved names in windows.

delta24:
@anantshri Ok, so we can remove that folder. Can we change the naming scheme for net and aux? Is it feasible?

anantshri:
Last time when i tried renaming the folder and all references inside somehow owtf was not working. So i suppose its feasable but it will just take some time.

as per discussion with @7a we decided to keep net as network and aux as auxillary.

delta24:
@anantshri Do you have to time to tackle this issue? I think you are best placed to do it :).

anantshri:
Like i keep saying i can give it a shot but not sure of a deadline or timeline when i will get it done.

delta24:
@anantshri  Sure no problem :)

delta24:
List of things to be done:

- [x]  Initial git clone problems - fuzzdb, aux->auxillary, net->network
- [ ] Research how to package the whole installer - NSI installer ? or simple installation like Kali, Debian?
- [ ] Powerscripts support

sachinkamath:
@delta24 I'd like to take this up. I am thinking of more like a simple installer (maybe a bat?). But the current problem is, all the installer scripts in the repo are written for Linux . Things like /dev/null and stuff will have to be re-written. 
The best thing to do'd be a separate branch for Windows. And why is this out of the OWTF Quality Release? Time issues?

sachinkamath:
[...]

PR #551, opened 2016-02-02
Closes issue #448 regarding owtf cloning in windows

DarKnight--:
This pull request has modified aux -> auxillary, net -> network and removed stale fuzzdb. Now Fuzzdb latest version is downloaded and used. New version does not have issue on windows.


anantshri:
I am away from my usual system but if this is working as expected. Consider a huge :+1: from my side. This is been a long standing issue #448

delta24:
@DarKnight--  There is still one error

error

delta24:
You now need to make a separate install file for Windows like Kali and Debian.

PR #552, opened 2016-02-06
Adding pref.sh from master to develop branch

C0smos:


delta24:
@C0smos : I meant the whole installation from the previous PR including ` install/debian/install.sh `, adding debian to ` install/install.py `, ` install/linux-distributions.cfg `.

C0smos:
Ah yeah my bad, I've added those files now

Issue #556, opened 2016-02-19
Listing plugins option (-l) not working


DoomTaper:
I tried command `./owtf.py -l net` and it shows error:
`sqlalchemy.exc.InvalidRequestError: Entity '<class 'framework.db.models.TestGroup'>' has no property 'type'`

delta24:
Merged in 4e45d761a911a7ebbdedf9648ddd66436113613d.

DePierre:
@delta24 at least test the PR. The fix is wrong.

@DoomTaper By commenting the line that raised an error, you are preventing filtering on plugin types. With your fix, instead of having 120 +/- 5 plugins, ``owtf.py -l web`` shows 600 +/- 30. Do you understand where that 500 hundreds extra plugins come from? Duplicates. Duplicates that were filtered out by the line you commented.

DePierre:
@DoomTaper is working on a proper fix.

DePierre:
@DoomTaper Thank you very much for that fix ;)

DoomTaper:
@DePierre credit goes to you as well :)

PR #567, opened 2016-03-03
Proper Fixes for issue #556

DoomTaper:
Its working properly now!

### DIFF
--- a/framework/core.py
+++ b/framework/core.py
@@ -294,6 +294,9 @@ class Core(BaseComponent):
                 # terminate()
                 self.TransactionLogger.poison_q.put('done')
                 self.TransactionLogger.join()
+            if getattr(self, "WorkerManager", None) is not None:
+                # Properly stop the workers.
+                self.WorkerManager.clean_up()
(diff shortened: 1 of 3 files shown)


delta24:
Tested - works ok now. @DePierre test once? :P

delta24:
Why is there an extra `[-]` in each category? 
https://gist.github.com/delta24/bc61bdf5a39fdd2ae87e

DoomTaper:
If I write `logging.info("message")` then output will be 
`[-] message`
but If I write `logging.info("\nmessage")` then output will be 
`[-] `
`message`
Similar case is happening above. Take a look [here](https://github.com/owtf/owtf/blob/develop/framework/plugin/plugin_handler.py#L476)
I hope you get it @delta24

delta24:
Ah, ok - I thought there was an empty entry from the file being parsed.

DePierre:
@DoomTaper Awesome thanks :)

DoomTaper:
My job!

PR #572, opened 2016-03-04
robots.txt spelling mistakes corrected

sharad1126:


DePierre:
@delta24 Silly question but... wasn't that on purpose?

PR #574, opened 2016-03-05
Ensures execution of Debian and Samurai scripts. Fixes #572

karandesai-96:
All the statements in shell scripts, which are meant to execute other shell scripts are prepended with "sh" in order to make sure that the script runs even if the permissions of script have the executable bit unset.

This workaround was enacted following the suggestion by @delta24 on #572 :
_"A possibly clean solution is to add either `sh` or `bash` in front of the script names everywhere. That way we don't play with permissions and source remains clean."_

### DIFF
--- a/install/aptitude-wrapper.sh
+++ b/install/aptitude-wrapper.sh
@@ -39,7 +39,7 @@ do
 done
 
 # Check if the current available disk size is enough.
-while [ "$(available_disk_size)" -lt "$size" ]; do
+while [ "${reset}$(available_disk_size)" -lt "$size" ]; do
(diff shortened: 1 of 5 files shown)


install/aptitude-wrapper.sh
```diff
@@ -39,7 +39,7 @@ do
 done
 
 # Check if the current available disk size is enough.
-while [ "$(available_disk_size)" -lt "$size" ]; do
+while [ "${reset}$(available_disk_size)" -lt "$size" ]; do
```
karandesai-96: Added ${reset} to make sure that both the strings are white, plain. Previous comparison was made between bold blue and normal white strings, hence throwing an Error.

PR #575, opened 2016-03-05
Fix #446: Improve Curl PUT test command output

### LINKED ISSUE #446: Improve curl commands
@7a : 

```bash
Tried File: 44b7fa5e6d0d5798
```

arunk-s:


DePierre:
@delta24 Could you have a look?

delta24:
@arunk-s Can you check the other `curl` commands in the `profiles/resources/default.cfg`?

arunk-s:
@delta24 Checked the other curl commands in the file. I've found that all of them should be easily repeatable as they involve (`OPTIONS, TRACE, DEBUG,`) checks. Also I've noticed that they all redirect `stderr` to `/dev/null` which is probably a good thing.

delta24:
Same as https://github.com/owtf/owtf/pull/600. Closing.

PR #576, opened 2016-03-05
filter by severity feature added

### LINKED ISSUE #453: Lack of filters on target page (UI)
It would be really nice if the UI offered a filter option on the target page:
There currently is a search, but this only searchs on the URL.
During very large assessments, it would be really handy to be able to filter by severity. For example:
Only show targets with "High" severity, only "Medium", etc.

saganshul:
@7a  @delta24 Please Review. Fixes #453
Screenshot attached with PR. 

screenshot from 2016-03-06 00-17-28

### DIFF
--- a/framework/interface/templates/target_manager.html
+++ b/framework/interface/templates/target_manager.html
@@ -62,18 +62,14 @@
                         </tr>
                     </thead>
                     <tbody class="searchable">
-                      <div class="dropdown">
-                        <button class="btn btn-primary dropdown-toggle" type="button" data-toggle="dropdown">
-                          Filter By Severity <span class="caret"></span>
-                        </button>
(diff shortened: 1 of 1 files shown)


DePierre:
@saganshul Wouldn't it be more user friendly to be able to select different ratings to show? Using radio buttons in the list for instance?

I could easily see someone wanting to only show Critical and High at the same time, while hidding the other ones.

saganshul:
@DePierre @7a @delta24 Filter made using checkbox as suggested. Please Review. Fixes #453  

screenshot from 2016-03-06 03-30-06
screenshot from 2016-03-06 03-30-10
screenshot from 2016-03-06 03-30-13

DePierre:
@saganshul Thanks for the new feature :) @7a will love it for sure!

PR #591, opened 2016-03-10
Fixed return value of get_config_value() function

pauTE:
Fixed problem with these scripts.I've tried to install owtf on my Kali Linux and the scripts were not creating the .owtf/proxy folder  and .owtf/db.cfg properly. That was  because the function get_config_value function (included in both scripts) was returning a bad value because the colours were making a bad path to the desired path.

And usage added to the /install/kali/install.sh script were teh OWTF project folder is required

### DIFF
--- a/install/kali/install.sh
+++ b/install/kali/install.sh
@@ -1,5 +1,11 @@
 #!/usr/bin/env sh
 
+
+usage="Usage:./install.sh OWTF_PROJECT_FOLDER"
+if [ $# -ne 1 ]; then
+    echo $usage
+    exit 1
+fi    
 # bring in the color variables: `normal`, `info`, `warning`, `danger`, `reset`
 . "$(dirname "$(readlink -f "$0")")/../colors.sh"
 
--- a/install/kali/install.sh
+++ b/install/kali/install.sh
@@ -1,11 +1,5 @@
 #!/usr/bin/env sh
 
-
-usage="Usage:./install.sh OWTF_PROJECT_FOLDER"
-if [ $# -ne 1 ]; then
-    echo $usage
-    exit 1
-fi    
 # bring in the color variables: `normal`, `info`, `warning`, `danger`, `reset`
 . "$(dirname "$(readlink -f "$0")")/../colors.sh"
 
--- a/install/db_config_setup.sh
+++ b/install/db_config_setup.sh
@@ -13,7 +13,7 @@ get_config_value(){
     parameter=$1
     file=$2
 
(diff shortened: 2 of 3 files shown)


karandesai-96:
I recently made the change with colorizing the logs. I first installed **OWTF** using the bootstrap script and both `./owtf/proxy` and `.owtf/db.cfg` files were made and stayed at their places since then, hence during subsequent runs I couldn't keep a track of the whole log as I ran the install script for many times during debugging. Thanks for catching this bug! :smile:

@delta24 I deleted these files manually, pulled this commit and tested it. It creates the files as it was meant to, works fine :smile:

delta24:
@pauTE Can you open a different pull request for each issue (in commits)? That way it is easier to track merges and revert whenever required.

pauTE:
Ok.No problem.




2016-03-12 7:07 GMT-05:00 Viyat Bhalodia <notifications@github.com>:

karandesai-96:
- Hello @pauTE you can even do it right now - it is very easy :smile: 
- I can share this nice trick I know with you - 
- Currently I assume you do not have any changes made locally after pushing the latest commit over here.
- You execute `git reset --soft HEAD^` This will delete the latest commit from the branch locally and you'll land up at the first commit of this branch, though the changes you made will still be left then and there - basically you'll reach to a state when you were about to make the second commit after all the changes were done to first commit.

- Use `git stash` and this will save the current state of your work at some separate place (your changes will disappear from the files here).

- Execute push to this branch using `--force` flag and the commit will be deleted from this branch on github.

- Checkout to `develop` branch locally and create a new branch. Then finally do a `git stash apply` to bring your changes there :smile: Push the new branch and create a separate PR..

Cool, isn't it ?! Cheers :grinning:

pauTE:
Ok I did it.Thank you so much @karandesai-96. @delta24

install/kali/install.sh
```diff
@@ -1,5 +1,11 @@
 #!/usr/bin/env sh
 
+
+usage="Usage:./install.sh OWTF_PROJECT_FOLDER"
+if [ $# -ne 1 ]; then
+    echo $usage
+    exit 1
+fi
```
delta24: Add a newline here.

pauTE:
@delta24 Done !!! :)

delta24:
Thank you @pauTE for the fix. :)

pauTE:
You are welcome.It's a pleasure. @delta24

2016-03-12 17:18 GMT+01:00 Viyat Bhalodia <notifications@github.com>:

PR #597, opened 2016-03-12
Fix #446 curl command improvement

### LINKED ISSUE #446: Improve curl commands

pauTE:
Hi, @delta24 ,I think that I may have a solution. The curl command has a flag ( -w) that is used to output some valuable information in order to compare a execution with another. In this case that we are using /dev/random , so we are getting a random input we can use this flag so we can compare the execution even with another that uses  the /dev/random.The http response code , the download time , the upload time , remote ip, effective url , are the kind of information that we can get.

So , for instance , with the following command:
 
`FILE=$(head /dev/random -c8 | od -tx1 -w16 | head -n1 | cut -d' ' -f2- | tr -d ' '); curl -s -i -A 'Mozilla/5.0 (X11; Linux i686; rv:6.0) Gecko/20100101 Firefox/15.0' -X 'PUT' -w \\n\\n\\nhttp-response-code:%{http_code}\\nRemote-IP:%{remote_ip}\\nRemote-Port%{remote_port}\\nUploaded-size:%{size_upload}\\nSize-Downloaded:%{size_download}\\nConnect-Time:%{time_connect}s\\nTotal-time:%{time_total}s\\nEffective-url:%{url_effective}\\n  --data "$FILE" -k "https://target.com/admin/$FILE" `

A part of the output will be:

http-response-code:501
Remote-IP:23.15.5.123
Remote-Port443
Uploaded-size:16
Size-Downloaded:258
Connect-Time:4.715s
Total-time:6.086s
Effective-url:https://target.com/admin/84e8c919333032da


delta24:
The fix is wrong because it calls` /dev/random` multiple times, see the end:

` "--data "$(head /dev/random -c8 | od -tx1 -w16 | head -n1 | cut -d' ' -f2- | tr  -d ' ')" -k "@@@target_url@@@/$(head /dev/random -c8 | od -tx1 -w6 | head -n1 | cut -d' ' -f2- | tr -d ' ')" 2> /dev/null"`

https://github.com/owtf/owtf/pull/575/files which stores the random result in a variable that is used The file should only be generated once, otherwise, we're pushing 1 file + trying to retrieve another, which will obviously always fail.

PR #598, opened 2016-03-13
Cleanup tests/run_tests.py and tests/runner/parser.py

karandesai-96:
I was going through certain files in "tests" directory - looking through the code for the trick which is causing all Travis builds to pass. I have still not found it but came across redundant code, memory overheads, avoidable interactions with recursion stack. Not sure that the code is still used or obsolete, but this PR is just to clean those files as I passed by reading them.

Commit body is used for description of the cleanup done :smile:

### DIFF
--- a/tests/run_tests.py
+++ b/tests/run_tests.py
@@ -1,33 +1,28 @@
 #!/usr/bin/env python
-'''
+"""
 This is the command-line tool for running tests.
-'''
-
+"""
+import os
+import sys
 from testing_framework.runner.parser import ArgumentParser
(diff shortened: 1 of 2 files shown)


karandesai-96:
I have cleaned up these two files which I had to refer for some understanding. Rest all I've seen so far only has PEP8 errors which are being rectified in some another PR. Nothing to add here from my side.

delta24:
@karandesai-96 I am closing this PR. Open another PR with

- [ ] Removing the existing tests
- [ ] Integrating the `owtf_testing` repo from @DePierre

PR #600, opened 2016-03-14
Fix #446 improve curl commands

### LINKED ISSUE #446: Improve curl commands

pauTE:
@delta24. Hi :)
As you said now I'm using a variable in order to call just once to /dev/random. All the valuable information is written in stdout as before. I hope this is the solution.

Thanks


delta24:
thank you @pauTE and @arunk-s  for the fix! :)

Issue #496, opened 2015-08-07
Error in handling special characters in url


DoomTaper:
If we write any special character in the url (lets say # ) "add target" button doen't work even after I refresh it.

![screenshot from 2015-08-07 19 35 52](https://cloud.githubusercontent.com/assets/10975138/9142537/4dec1e98-3d0e-11e5-926c-11bf6dd818f7.png)

![screenshot from 2015-08-07 19 36 07](https://cloud.githubusercontent.com/assets/10975138/9142547/63f48446-3d0e-11e5-93ce-34457dfc7909.png)

DePierre:
@DoomTaper Hi there. I tried to add the target 'http://example#.com' and the 'Add Targets' button is working fine. It still works after trying to add the target, when refreshing the page, when switching back and forth between tabs.

Have you install the develop branch? Could you let us know what commit you are at? You can find the hash by ``cd``ing into ~/owtf and execute ``git log -1``.

DoomTaper:
I was not on develop branch. After installing devlop branch as i run ./owtf.py I got the error stated below



Traceback (most recent call last):
  File "./owtf.py", line 300, in <module>
    main(sys.argv)
  File "./owtf.py", line 263, in main
    ComponentInitialiser.initialisation_phase_1(root_dir, owtf_pid)
  File "/root/Desktop/owtf_develop/owtf/framework/dependency_management/component_initialiser.py", line 69, in initialisation_phase_1
    PluginDB()
  File "/root/Desktop/owtf_develop/owtf/framework/db/plugin_manager.py", line 24, in __init__
    self.LoadWebTestGroups(self.config.FrameworkConfigGet("WEB_TEST_GROUPS"))
  File "/root/Desktop/owtf_develop/owtf/framework/db/plugin_manager.py", line 56, in LoadWebTestGroups
    group="web")
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 1613, in merge
    load=load, _recursive=_recursive)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/session.py", line 1652, in _merge
    merged = self.query(mapper.class_).get(key[1])
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/query.py", line 820, in get
    return loading.load_on_ident(self, key)
  File "/usr/local/lib/python2.7/dist-packages/sqlalchemy/orm/loading.py", line 231, in load_on_ident
    return q.one()
[...]

Issue #605, opened 2016-03-17
UI Add Targets button bug


RahulPratapSingh:
Add Targets button get disabled even if the url is not valid.
![screenshot from 2016-03-17 12-18-13](https://cloud.githubusercontent.com/assets/8836104/13843945/b1a8d40a-ec5c-11e5-9c4a-d2bfe66863f0.png)

DoomTaper:
@RahulPratapSingh are you on develop branch ?
Seems similar to issue #496

RahulPratapSingh:
@DoomTaper yes. I am on develop branch. This bug seems similar to bug #496. I checked that now, but this issue isn't addressed in code. Even, to replicate no need to put any special character, just leave it blank and hit Add Targets!. It will get disabled and will remain same, until user refresh the page or do any other activity. It's seems to be annoying to user to refresh page everytime.

sachinkamath:
@RahulPratapSingh I am not able to reproduce it. It works as intended.

Typed an Invalid URL first, also tried by leaving it blank : 
![1](https://cloud.githubusercontent.com/assets/9546091/13851057/a4abbb48-ec82-11e5-9c53-89a075d1a1ea.png)


Used a legit URL (https://www.google.com) and clicked again : 

![2](https://cloud.githubusercontent.com/assets/9546091/13851062/a9f177fa-ec82-11e5-801a-ecb406a85148.png)

RahulPratapSingh:
@sachinkamath My bad. I checked it again before posting, but still compared with old (master branch code). I will make sure, this blunders won't happen again. :(

PR #606, opened 2016-03-17
UI add targets button bug fixed

RahulPratapSingh:
UI Add Targets button gets disabled even if the url is not valid. Enabled the add targets button, when url is not valid. Addresses bug #496


sachinkamath:
@RahulPratapSingh I guess you can close this as #605 is closed.

PR #610, opened 2016-03-17
Fixed Reflected XSS Vulnerability and Removed Redundant Username and Password

RahulPratapSingh:
Fixed Reflected XSS Vulnerability
Redundant username and password in dictionary list are removed.


sachinkamath:
@RahulPratapSingh I think it's a bad idea to mix up your commits. @delta24 would probably agree.

PR #611, opened 2016-03-17
Fixed session deletion bug

### LINKED ISSUE #608: Unexpected Behavior after deleting a session
#### Steps to Reproduce : 

1) Navigate to Targets, click on Sessions and add a New Session, say `Test Session`.

screenshot from 2016-03-17 23-04-50
<hr>
2) Select `Test session` as the session and then delete it, Click outside the popup to close it.

screenshot from 2016-03-17 23-05-51
<hr>

3) Notice that Test Session is still selected as the session. Try adding a Target URL. You'll be greeted with an error.

screenshot from 2016-03-17 23-06-26
<hr>

#### What's happening?####

As far as I understood, the `delete` method in `OWTFSessionHandler` deletes the session from the `db` but doesn't set it to the default value after deletion.

Snippet from api_handlers.py : 
```python 
def delete(self, session_id=None, action=None):
        if session_id is None or \
           action is not None:
        try:
            self.get_component("session_db").delete_session(
                int(session_id))
        except exceptions.InvalidSessionReference:
```

sachinkamath:
Fixes #608


framework/interface/api_handlers.py
```diff
@@ -166,6 +166,11 @@ def delete(self, session_id=None, action=None):
         try:
             self.get_component("session_db").delete_session(
                 int(session_id))
+            # gets the id of the default session. 
+            default_session = self.get_component("session_db").get_default_id()
```
sachinkamath: The default session's ID is after deletion because I noticed that when I try to delete the default session from the UI, it is deleted and a new "default session" is written into the database.

delta24:
I think the default session will always be there if there are no other sessions.

sachinkamath:
@delta24 Yes. Please check the issue #608 and then check this. I closed this becuase I had pulled the UI changes. Will make a fresh one

Issue #607, opened 2016-03-17
Verify distribution during installation


sachinkamath:
I was trying to install OWTF onto an Ubuntu VM today and I *accidentally* chose Kali as my distribution.  The installation went fine and I was able to fire up the web UI just fine. But as expected, there were a lot of errors when I tested it by scanning a website like this as nikto was not installed : 

![screenshot from 2016-03-17 21-18-43](https://cloud.githubusercontent.com/assets/9546091/13851905/f2355aa6-ec85-11e5-9ded-9a08c8bdaf38.png)

Some of the things that can be done to resolve this : 
* Instead of prompting user's to select their distribution, dynamically detect it using `lsb_release -a`
* Verify if the dependencies are installed before firing up OWTF (or) add a button in the WebGUI to verify the dependencies.

delta24:
We already detect the distribution automatically and the dependencies. @sachinkamath see the `install.py` and the `dependency_management` dir.

sachinkamath:
@delta24 Gotcha. I had discussed about this on IRC with @7a . We concluded that auto-detection of the OS will help. I'll send a PR soon for that :)

PR #618, opened 2016-03-18
Enhanced installer script 

sachinkamath:
This should be a fix to #607 I have tested it on Ubuntu, Fedora and Kali Linux. 

The script tries to detect an instance of `apt-get` and assigns the Debian tag to the OS if is found. If `apt-get` is not found. it will raise an error and ask the user if he wants to proceed as discussed with @7a and @ner0x652 on IRC. This way, any user who has `apt` will now be tagged as a Debian user and the installation will continue automatically (Tested on Ubuntu) . 

Screenshot of Ubuntu : 
screenshot from 2016-03-18 16-35-33

A user without `apt` will see screens like this (Screenshots of Fedora) : 
fedora1
If the user wishes not to continue, he gets the following message : 
fedora2

If the user wishes to continue, he is given the selection screen for the OS (I'm doing this to eliminate the edge cases, where a user might have installed a spin/derivative of an OS mentioned there)

fedora3

After selection of the OS, the installation continues as expected (for the brave user :man_with_gua_pi_mao: )

fedora4

It should auto-detect Kali fine as before :)

**Edit:** 

Tested on Kali as well. Here you go : 
screenshot from 2016-03-19 18 03 23

### DIFF
--- a/install/install.py
+++ b/install/install.py
@@ -26,6 +26,22 @@ class Installer(object):
         self.parser.add_argument('--no-user-input', help='run script with default options for user input', action="store_true")
         self.parser.add_argument('--core-only', help='install only owtf dependencies, skip optional tools', action="store_true")
 
+    def print_success(self, error_code):
+        if(error_code == 1):
+            Colorizer.danger("\n[!] The installation was not successful.")
+            Colorizer.normal("[*] Visit https://github.com/owtf/owtf for help ")
+        else:
+            Colorizer.success("[*] Finished!")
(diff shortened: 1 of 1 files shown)


sachinkamath:
@delta24 @7a @DePierre Please review.

delta24:
This is great but send a clean PR with only the relevant commits. Also avoid the fast-forwarded merges like https://github.com/sachinkamath/owtf/commit/6e10f1086453eebb5520e201613904a3f46ec46b - they make the git commit history dirty.

sachinkamath:
Made a new PR #627

PR #622, opened 2016-03-19
Fixed session deletion bug

### LINKED ISSUE #608: Unexpected Behavior after deleting a session

sachinkamath:
This fixes #608 . 

Solution implemented : 
After deletion, switch the session to the default session.


framework/db/session_manager.py
```diff
@@ -53,6 +53,12 @@ def get_session_id(self):
             models.Session.id).filter_by(active=True).first()
         return session_id
 
+    # returns the session ID of the default session
+    def get_default_id(self):
+        session_obj = self.db.session.query(
+            models.Session).filter_by(name="default session").first()
```
delta24: Fix PEP8 120 char limit.

framework/interface/api_handlers.py
```diff
@@ -166,6 +166,13 @@ def delete(self, session_id=None, action=None):
         try:
             self.get_component("session_db").delete_session(
                 int(session_id))
+            # gets the id of the default session. 
+            default_session = self.get_component("session_db").get_default_id()
+            # sets the session id as the default session's one
+            self.get_component("session_db").set_session(
```
delta24: Same here.

delta24:
@sachinkamath, can you squash your commits in one or send a clean PR? The change is small and does not need the 4 commits in the PR :) .

PR #629, opened 2016-03-21
Added installation scripts

sachinkamath:
@delta24  Done :) Tested on Fedora Kali and Ubuntu. Works as in screenshots of pull #618


install/install.py
```diff
@@ -113,6 +129,16 @@ def install(cmd_arguments):
...
     cp.read(distros_cfg)
 
+    # Check if the current distribution is compatible 
+    if not is_compatible():
+        Colorizer.danger("\n[!] You don't seem to have the apt package manager installed. "
+                "This means that your distribution is not fully supported (yet).")
+
+        # yes is assumed for any other input
+        response = raw_input("\nContinue installation anyway? [Y/n] : ")
+        if(response.lower() == 'n'):
```
delta24: Parenthesis again.

install/install.py
```diff
@@ -113,6 +129,16 @@ def install(cmd_arguments):
     cp = ConfigParser.ConfigParser({"RootDir": root_dir, "Pid": pid})
     cp.read(distros_cfg)
 
+    # Check if the current distribution is compatible 
+    if not is_compatible():
```
delta24: This is not needed. We do support non-Debian based distro (Archlinux) and will support other distros. So remove this check.

install/install.py
```diff
@@ -103,6 +103,22 @@ def install_restricted_from_cfg(config_file):
         install_in_directory(os.path.expanduser(cp.get(section, "directory")), cp.get(section, "command"))
 
 
+def is_compatible():
+        compatible_value = os.system("which apt-get >> /dev/null")
```
delta24: If the distro is not Debian based, it outputs an error on the screen (from `which`)

```
which: no apt-get in $PATH
```
which looks like there was an error in OWTF installer.

Instead suppress that error using `2>/dev/null` or `2>&1`.

PR #631, opened 2016-03-23
Fixed github.com redirection urls

RahulPratapSingh:
Some github.com urls are redirecting. Fixed them.

1) https://api.github.com/repos/owtf/owtf-zest-jars/tarball
302 redirection

2) https://github.com/fuzzdb-project/fuzzdb/archive/master.zip
302 redirection

3) https://raw.github.com/pypa/pip/master/contrib/get-pip.py
301 redirection

### DIFF
--- a/install/debian/install.sh
+++ b/install/debian/install.sh
@@ -23,7 +23,7 @@ sudo apt-get update
 
 # Grab and install pip
 echo "${info}[*] Installing pip using get-pip.py${reset}"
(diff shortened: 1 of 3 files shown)


delta24:
I am closing this PR for 2 reasons:

- The PR is not in the format the template has specified.
- The changes done are cosmetic only. The 301 redirections are there because Github changes its internals frequently and we should not have to update the redirections every time they change something (in this case - what if the API breaks?).

PR #634, opened 2016-03-26
Prevents session manager from closing without a value

sachinkamath:
This PR prevents the `SessionModal` from closing unless a session is selected.
## Description

A couple of parameters were introduced into the HTML code that prevented mouse-click outside of the modal and Escape key from closing the modal
## Related Issue

This is a fix for Issue #608 

## Reviewers

@delta24 
## How Has This Been Tested?

This was tested on IceWeasel, Firefox and Chromium and works as intended on all the three browsers. The `SessionModal` does not close unless a session is selected by the user.

## Types of changes

- [ ] Other ()

### Checklist:

### DIFF
--- a/framework/interface/templates/target_manager.html
+++ b/framework/interface/templates/target_manager.html
@@ -86,7 +86,7 @@
     worklist_post_func="postToWorkList") %}
 
 <!-- Small modal -->
-<div class="modal fade" id="sessionModal" tabindex="-1" role="dialog" aria-labelledby="sessionModalLabel" aria-hidden="true">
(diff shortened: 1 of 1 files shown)


delta24:
@sachinkamath Your fix is invalid and breaks other functionality. Try adding a new session and modal cannot be closed unless you click on the **selected** radio button of the new session.

A correct fix would be a function which checks that `checked = (obj.active)'';` returns `True` after deleting a session (i.e. the user has selected another session) - only then allow the modal to be closed.

sachinkamath:
@delta24 not allowing the modal to close when the user had not made a selection seems to be a bad idea (Please check IRC regarding this) . Instead, I'll add a function that will warn the user that a session is not selected when he exits the modal without selecting a session and enforce the same when the user tries to add target without a valid session. The best way to do this is probably by using  `/api/sessions`

Issue #635, opened 2016-03-26
Session Modal breaks for large session names


sachinkamath:
`SessionModal` handles the sessions in `/ui/targets` . When a large string is passed as session name, it stretches beyond the size of modal window and breaks the delete functionality
## Expected Behavior

The session name should have been truncated.
## Current Behavior 

The Session Name overflows out of bounds causing delete button to disappear
## Possible Solution

* Truncate the session name / Set a limit (Better Fix)
(or)
* Increase the modal size

## Steps to Reproduce (for bugs)

1. Open the Sessions and get the Modal popup.
2. Enter a long string.

![screenshot from 2016-03-26 11 40 57](https://cloud.githubusercontent.com/assets/9546091/14058477/09f6e632-f2f8-11e5-8b89-fa8d5fb46f6f.png)

## Your Environment

* Installation method used: Manual
* Version/branch used: develop
* Operating System and version (like Kali, Debian, ArchLinux, etc): Kali Linux
* pip/setuptools version: 8.1.1
* sudo access: Yes

delta24:
@sachinkamath @DePierre @tunnelshade @7a: I think better to truncate the session names to max. 50 chars.

tunnelshade:
@delta24 Yes, put a limit on it. 50 characters make sense

delta24:
@tunnelshade should we also start implementing length limits in the sqlalchemy models itself?

tunnelshade:
Yes @delta24 . Then catch for exceptions while creating models.

delta24:
Closed in https://github.com/owtf/owtf/commit/d9a2cf39cf96f792adc8c9558953931a01361b72.

PR #636, opened 2016-03-26
Truncated long session names

sachinkamath:
This PR truncates the length of session names to 50 characters.

## Description

Ref : #635 
The SessionModal breaks when really long session names were given as input.
## Related Issue

Issue : #635 

## Reviewers

@delta24 
## How Has This Been Tested?

Manually tested on a browser.

screenshot from 2016-03-26 12-27-13

## Types of changes

- [ ] Other ()

### Checklist:

### DIFF
--- a/framework/interface/templates/target_manager.html
+++ b/framework/interface/templates/target_manager.html
@@ -180,6 +180,7 @@ function removeTargetFromCurrentSession(target_id) {
 // New session
 function newSession(session_name) {
     if (session_name) {
+        session_name = session_name.substring(0, 50);
         $.ajax({
                 url:mySpace.owtf_sessions_api_url,
                 type:'POST',
--- a/framework/db/session_manager.py
+++ b/framework/db/session_manager.py
@@ -57,7 +57,7 @@ class OWTFSessionDB(BaseComponent):
         existing_obj = self.db.session.query(
             models.Session).filter_by(name=session_name).first()
         if existing_obj is None:
-            session_obj = models.Session(name=session_name)
+            session_obj = models.Session(name=session_name[:50])
             self.db.session.add(session_obj)
             self.db.session.commit()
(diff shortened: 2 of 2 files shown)


framework/interface/templates/target_manager.html
```diff
@@ -93,7 +93,7 @@
             <div class="input-group">
                 <input id="newSessionName" type="text" class="form-control" placeholder="New Session">
                 <span class="input-group-btn">
-                    <button class="btn btn-primary" onclick="newSession($('#newSessionName').val());" type="button">Add!</button>
+                    <button class="btn btn-primary" onclick="newSession($('#newSessionName').val().substring(0,50));" type="button">Add!</button>
```
delta24: `substring(0, 50)`

delta24:
@sachinkamath what if someone intercepts that request and change the length again? You are only restricting the length when adding the session.

sachinkamath:
Very good catch :) I didn't think about that. I will update the PR soon

sachinkamath:
@delta24 Fixed :)

delta24:
@sachinkamath look at the comments on the corresponding issue for the update.

framework/db/models.py
```diff
@@ -25,7 +25,7 @@ class Session(Base):
     __tablename__ = "sessions"
 
     id = Column(Integer, primary_key=True, autoincrement=True)
-    name = Column(String, unique=True)
+    name = Column(String(50), unique=True)
```
delta24: In this case, if I bypass the ajax request and still have a session name longer than 50 char, then this will result in an uncaught exception and sqlalchemy will probably crash. Instead, truncate the length here,  https://github.com/owtf/owtf/blob/develop/framework/db/session_manager.py#L60:
'
`session_obj = models.Session(name=session_name[:50])`

PR #641, opened 2016-03-27
Session deletion fix

sachinkamath:
This PR explicitly reports errors that might occur while a session is not selected. Ref : #608 
## Description

* A function was added to check if an active session exists using API calls to `/api/sessions/`
* Beautified JS

## Related Issue

This is a fix for Issue #608 

## Reviewers

@delta24 @DePierre @7a @tunnelshade 
## How Has This Been Tested?

This was tested on IceWeasel, Firefox and Chromium and works as intended on all the three browsers. 

## Types of changes

- [ ] Other ()

### Checklist:

### DIFF
--- a/framework/interface/templates/target_manager.html
+++ b/framework/interface/templates/target_manager.html
@@ -93,7 +93,7 @@
             <div class="input-group">
                 <input id="newSessionName" type="text" class="form-control" placeholder="New Session">
                 <span class="input-group-btn">
-                    <button class="btn btn-primary" onclick="newSession($('#newSessionName').val());" type="button">Add!</button>
+                    <button class="btn btn-primary" onclick="newSession($('#newSessionName').val().substring(0, 50));" type="button">Add!</button>
                 </span>
             </div><!-- /input-group -->
         </div>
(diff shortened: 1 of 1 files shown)


mikefitz888:
Hi,
The beautification looks nice. There's one thing that needs to be fixed and I have a few optional suggestions.

To Fix:
There's a redundancy on lines 210-212, calling: 
```
updateSessions();
updateCurrentSessionInfo();
```

Then calling them both again inside (line 255):
`checkForActiveSession(1);`

Optional:
A for-loop can be used instead of jQuery's `.each()` which will be faster and allows you to `break` out of it if you find an active session.

[https://jsperf.com/browser-diet-jquery-each-vs-for-loop/123](url)
Comparison of the speed of various for-each approaches.

Additionally you could make the `checkForActiveSession` function more flexible by passing a 'callback function' to it instead of an alertLevel. Here's an example of the code (with both optional features) and a sample use case:

```
function checkForActiveSession(callback) {
    updateSessions();
    updateCurrentSessionInfo();
    $.ajax({
        url: '/api/sessions/',
        type: 'GET',
        cache: false,
        success: function(data) {
            var active_session_exists = false;
            for(var i = 0; i < data.length; i++){
                if(data[i]["active"]) {
                    active_session_exists = true;
                    break;
                }
            }
            callback(active_session_exists);
        },
        error: function (xhr, textStatus, serverResponse) {
            alertFail("Unable to get sessions");
        }
    });	
}
```
Example usage:
```
checkForActiveSession(function(session_exists){
    if(session_exists){
        //do something when active session exists
    } else {
        //do something when no active session available
    }  
});
```

delta24:
@sachinkamath Please fix the PR using @mikefitz888 suggestion. :)

delta24:
@sachinkamath IMO a better fix is to disable the modal collapse (by clicking in the background) until the user chooses a session. Refer to the Bootstrap docs for this.

DePierre:
@sachinkamath @delta24 What is the current status of this PR?

delta24:
@DePierre Right now the fix allows the modal window to be closed but it displays a Bootstrap alert saying you do not have any session selected. IMO the best fix would be to disable modal collapse by clicking in the background (there is a certain `data-backdrop` that allows it I think) - until the user selects a particular session.

delta24:
Closing this as the fix is not correct.

PR #653, opened 2016-04-03
Progress Bar Bug Fixed. Fixes #644

### LINKED ISSUE #644: Bug in progress bar
The major problem lies in the [formula](https://github.com/owtf/owtf/blob/develop/framework/interface/templates/manager_interface.html#L180) of calculating the percentage of completed work i.e. `complete_count / (left_count + complete_count)`
There are **two bugs** related to progress bar.
#### First bug
1. run `./scripts/db_setup.sh clean` and then `./scripts/db_setup.sh init` to ensure your db is clean.
2. Now add a target and run a single plugin against it. Now according to formula while the plugin is running `left_count=0` because work has been taken and `complete_count=0` because work is not finished yet. So in the formula `complete_count / (left_count + complete_count)` it will be 0/0 which is an error although it will not showup anywhere and that's why progress bar doesn't work (don't think progress bar was working for not showing any work , read my next bug to get clear about this).

#### Second bug
1. make sure you have successfully run a plugin so that `complete_count != 0` now run a **single plugin** against a target ( try to select a plugin which takes time like i tested for arachni_unauth web/active ). 
2. Go to the worklist page and you will see that progress bar is showing 100% completion even if plugin is running (see the terminal) because in this case `left_count = 0` (because work has been taken) so the formula `complete_count / (left_count + complete_count) ` becomes 1 ( remember `complete_count != 0`) but if `complete_count = 0` as in previous case rather than showing 100% progress bar will not work instead.

**P.S.**: Sorry to the collaborators for not following the template but this was not fitting into it :P

saganshul:
This PR fixes the progress bar bugs referenced in #644

## Description
Fix of Bug-1 : To fix this bug, PR checks left_count == 0 and complete count == 0 and if it is true, It just replaces the progress bar with simple message.
Fix of Bug-2 : To fix this bug, PR increases the left_count by 1 because what was happening previously that if the plugin was running it assumed to be completed as it is was removed from work-list. Also at condition where left_count=1. PR checks the busy attribute of every worker and if no worker is running PR decreases the left_count by 1. Hence this fixes Bug-2
  
## Related Issue
#644 

## Reviewers
@delta24 @DePierre @DoomTaper @7a 

## How Has This Been Tested?
Tested on Firefox and Chrome, working on both the browsers

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/framework/interface/templates/manager_interface.html
+++ b/framework/interface/templates/manager_interface.html
@@ -232,6 +232,7 @@ function moveProgressBar() {
                 elem.css("width", percentage_done + '%');
                 if (percentage_done == 100) {
                     elem.removeClass('active');
+                    $('#progressbar').html('<p>Worklist is empty</p>');
                     clearInterval(progressbar);
                 }
             }
--- a/framework/interface/templates/manager_interface.html
+++ b/framework/interface/templates/manager_interface.html
@@ -19,9 +19,11 @@
     </div>
 </div>
 <br>
-<h5><strong>Progress of Running Plugins</strong></h5>
-<div class="progress">
-  <div id="pluginprogress" class="progress-bar progress-bar-success progress-bar-striped active" role="progressbar" aria-valuenow="0" aria-valuemin="0" aria-valuemax="100" style="width: 0%">
+<h5><strong>Total scan progress</strong></h5>
+<div id="progressbar">
+  <div class="progress">
+    <div id="pluginprogress" class="progress-bar progress-bar-success progress-bar-striped active" role="progressbar" aria-valuenow="" aria-valuemin="0" aria-valuemax="100" style="width: 0%">
+    </div>
   </div>
 </div>
 <div class="row" id="workersInfo">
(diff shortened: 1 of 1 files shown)


DoomTaper:
@saganshul there is still a problem, you are using `left_count++` which means you assumed that there will always be single worker working what if there are multiple workers working at the same time (say 3 workers) then you need to increase you left_count by 3 rather than 1.
**Solution**: A possible solution is to check workers' status in every interval and increase left_count accordingly not just by 1.What do you think @delta24 @7a @DePierre ?

saganshul:
@DoomTaper , It is good idea we can easily do that. Well thank you for pointing out that.
@delta24 @DePierre @7a your suggestions please

saganshul:
@delta24 @DePierre @DoomTaper PR updated. Please review.

delta24:
The progress bar is still not very intuitive. For example, if the scan is completed and there are no plugins in the worklist - the progress bar should show something other than a full bar otherwise it looks like the progress has been stopped (or scanning is stuck).

scrot_04-11-2016_12-56-37

saganshul:
@delta24 required changes made, please review.

DoomTaper:
@delta24 it seems fine to me.

PR #672, opened 2016-04-21
Fixes #668

### LINKED ISSUE #668: PostgreSQL Fix in db_setup.sh restarts postgresql daemon in any case
The fix for postgresql SSL configuration in the file ``db_setup.sh`` restarts the daemon even in the case when the user said "no" to the fix:

```bash
$ ./scripts/db_setup.sh init 

------------------------- OWTF Database Helper Script -------------------------
                    Helps in creation of user and database                     
-------------------------------------------------------------------------------

[+] Ensure that you have required values in /root/.owtf/db.cfg

Press Enter to continue
n
CREATE ROLE
CREATE DATABASE
Having SSL=true in postgres config causes many errors (psycopg2 problem)
Remove SSL=true from the PostgreSQL config?[Y/n]n
Restarting the postgresql service
```

https://github.com/owtf/owtf/blob/develop/scripts/db_setup.sh#L33 does not check the user input before restarting the daemon.

I would expect the script to do nothing if I answer "no", especially since restarting the daemon takes quite some time.

DoomTaper:
## Description

Section which was restaring the postgresql service has now moved to the "yes" case. So if user doesn't answer yes then it will not restart postgresql.
## Related Issue

#668 

## Reviewers

@DePierre 

## How Has This Been Tested?

running `./scripts/db_setup.sh init` and then if user doesn't answers "yes" then it will not restart the postgresql.

Before
screenshot 88

After
screenshot 89

## Types of changes

- [ ] Other

### DIFF
--- a/scripts/db_setup.sh
+++ b/scripts/db_setup.sh
@@ -41,27 +41,26 @@ postgresql_fix() {
   case $remove_ssl in
     [yY][eE][sS]|[yY])
       sed -i -e '/ssl =/ s/= .*/= false/' $postgres_conf
+      echo "Restarting the postgresql service"
+      service_bin=$(which service | wc -l)
+      systemctl_bin=$(which systemctl | wc -l)
+      if [ "$service_bin" = "1" ]; then
+        service postgresql restart
+        service postgresql status | grep -q '^Running clusters: ..*$'
+        status_exitcode="$?"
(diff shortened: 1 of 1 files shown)


DoomTaper:
@delta24 fixes the issue just before creating this PR so closing this.

PR #673, opened 2016-04-21
Added file check before reading config files, Fixes #666

### LINKED ISSUE #666: ConfigDB silently fails when default.cfg not found
When running OWTF from a different directory and OWTF cannot find the file `profile/general/default.cfg`, it silently fails.

Instead of throwing a verbose error message such as "Configuration file not found", it simply does not load the configuration DB values and later fails when loading Zest with the following error message:
```bash
Traceback (most recent call last):
  File "/home/depierre/git-repo/owtf/owtf_testing/tests_functional/cli/test_empty_run.py", line 13, in test_cli_empty_run
    self.run_owtf()
  File "/home/depierre/git-repo/owtf/owtf_testing/utils/owtftest.py", line 55, in run_owtf
    owtf.main(args)
  File "/home/depierre/git-repo/owtf/owtf.py", line 235, in main
    ComponentInitialiser.initialisation_phase_1(root_dir, owtf_pid)
  File "/home/depierre/git-repo/owtf/framework/dependency_management/component_initialiser.py", line 73, in initialisation_phase_1
    zest.init()
  File "/home/depierre/git-repo/owtf/framework/zest.py", line 25, in init
    self.StopRecorder()  # recorded should be stopped when OWTF starts
  File "/home/depierre/git-repo/owtf/framework/zest.py", line 145, in StopRecorder
    self.db_config.Update("ZEST_RECORDING", "False")
  File "/home/depierre/git-repo/owtf/framework/db/config_manager.py", line 114, in Update
    raise InvalidConfigurationReference("No setting exists with key: " + str(key))
InvalidConfigurationReference: 'No setting exists with key: ZEST_RECORDING'
```

The error is in https://github.com/owtf/owtf/blob/develop/framework/db/config_manager.py#L31 where the documentation of `RawConfigParser.read()` says:

> If none of the named files exist, the ConfigParser instance will contain an empty dataset.

To fix the issue and get a verbose message, the function `LoadConfigDBFromFile` should use https://docs.python.org/2/library/configparser.html#ConfigParser.RawConfigParser.readfp.
Another approach would be to check that `file_path` exists before https://github.com/owtf/owtf/blob/develop/framework/db/config_manager.py#L31

DarKnight--:
The config files will be checked for existence before reading it, to prevent exception.

## Related Issue

issue #666 

## Reviewers

@delta24 @DePierre 

## How Has This Been Tested?

Delete the config file in `profile` folder and try to run OWTF

screenshot from 2016-04-21 19 24 20

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/framework/db/config_manager.py
+++ b/framework/db/config_manager.py
@@ -5,6 +5,7 @@ from framework.db import models
 from framework.lib.general import cprint
 import ConfigParser
 import logging
+import os
 
 
 class ConfigDB(BaseComponent, DBConfigInterface):
@@ -15,6 +16,7 @@ class ConfigDB(BaseComponent, DBConfigInterface):
         self.register_in_service_locator()
         self.config = self.get_component("config")
         self.db = self.get_component("db")
+        self.error_handler = self.get_component("error_handler")
         self.LoadConfigDBFromFile(self.config.get_profile_path('GENERAL_PROFILE'))
 
     def IsConvertable(self, value, conv):
@@ -28,6 +30,8 @@ class ConfigDB(BaseComponent, DBConfigInterface):
         logging.info("Loading Configuration from: " + file_path + " ..")
         config_parser = ConfigParser.RawConfigParser()
         config_parser.optionxform = str  # Otherwise all the keys are converted to lowercase xD
+        if not os.path.isfile(file_path):  # check if the config file exists
+            self.error_handler.FrameworkAbort("Config file not found at: %s" % file_path)
         config_parser.read(file_path)
         for section in config_parser.sections():
             for key, value in config_parser.items(section):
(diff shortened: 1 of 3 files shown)


DePierre:
Would it be better to, instead of relying on db error handler, get the component directly?

For instance something like:
```python
class ConfigDB(BaseComponent, DBConfigInterface):

    COMPONENT_NAME = "db_config"

    def __init__(self):
         # . . .
        self.error_handler = self.get_component("error_handler")
    # . . .
    def LoadConfigDBFromFile(self, file_path):
        # . . .
        if not os.path.isfile(file_path):  # Check if the config file exists.
            self.error_handler.FrameworkAbort("Config file not found at: %s" % file_path)
```

Also could you make sure you follow pep8? With the missing surrounding spaces and the proper inline comment format.

DePierre:
@DarKnight-- Thank you very much! :)

PR #682, opened 2016-05-03
Addded plugin name to code converter function Fixes #650

### LINKED ISSUE #650: Plugin not running if plugin_name is given from CLI
Plugin is not running if I gave plugin name instead of plugin code with -o argument through CLI.

## Expected Behavior

OWTF should run the plugin with plugin name also if given from the CLI.

## Current Behavior

If plugin name is given with -o argument then it doesn't run

## Possible Solution

IMO in `GetAll()` function ( [here](https://github.com/owtf/owtf/blob/develop/framework/db/plugin_manager.py#L215) ) criteria is not checked if it contains plugin name or not. So, need to fix that only I guess.

Run the following command and see the terminal 
`./owtf.py -o Old_Backup_and_Unreferenced_Files https://example.org`

## Context

Plugins will be able to run from their names also.

## Your Environment

* Installation method used: automatic
* Version/branch used:develop
* Operating System and version (like Kali, Debian, ArchLinux, etc): kali
* pip/setuptools version: 8.0.3
* sudo access: yes

DoomTaper:
## Description

Added plugin name to code converter function so that before querying all plugin names are converted to their respective codes

## Related Issue

#650 

## Reviewers

@delta24 @DePierre 

## How Has This Been Tested?

tested command `./owtf.py -o Application_Discovery http://127.0.0.1`

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/framework/db/plugin_manager.py
+++ b/framework/db/plugin_manager.py
@@ -212,7 +212,17 @@ class PluginDB(BaseComponent, DBPluginInterface):
                 query = query.filter(models.Plugin.name.in_(criteria["name"]))
         return query.order_by(models.TestGroup.priority.desc())
 
+    def PluginNametoCode(self, codes):
+        query = self.db.session.query(models.Plugin.code)
+        for count, name in enumerate(codes):
+            if "OWTF-" not in name:
+                code = query.filter(models.Plugin.name == name).first()
+                codes[count] = str(code[0])
+        return(codes)
+
     def GetAll(self, Criteria={}):
+        if Criteria.has_key("code"):
+            Criteria["code"] = self.PluginNametoCode(Criteria["code"])
         query = self.GenerateQueryUsingSession(Criteria)
         plugin_obj_list = query.all()
         return(self.DerivePluginDicts(plugin_obj_list))


framework/db/plugin_manager.py
```diff
@@ -212,7 +212,18 @@ def GenerateQueryUsingSession(self, criteria):
...
 
+    def PluginNametoCode(self, codes):
+        query = self.db.session.query(models.Plugin.code)
+        for i in range(len(codes)):
+            if "OWTF-" not in codes[i]:
+                cd = query.filter(models.Plugin.name == codes[i]).first()
+                codes[i] = str(cd[0])
+        return(codes)
+
+
```
DarKnight--: IMO 1 line space should be there.

framework/db/plugin_manager.py
```diff
@@ -212,7 +212,18 @@ def GenerateQueryUsingSession(self, criteria):
                 query = query.filter(models.Plugin.name.in_(criteria["name"]))
         return query.order_by(models.TestGroup.priority.desc())
 
+    def PluginNametoCode(self, codes):
+        query = self.db.session.query(models.Plugin.code)
+        for i in range(len(codes)):
```
delta24: Use `enumerate` instead of  `range(len(xx))`.

PR #684, opened 2016-05-05
Revert "Addded plugin name to code converter function Fixes #650"

### LINKED ISSUE #650: Plugin not running if plugin_name is given from CLI

DoomTaper:
Reverts owtf/owtf#682. Issue described by @DarKnight-- is legit, needs to revert this asap.

### DIFF
--- a/framework/db/plugin_manager.py
+++ b/framework/db/plugin_manager.py
@@ -214,17 +214,7 @@ class PluginDB(BaseComponent, DBPluginInterface):
                 query = query.filter(models.Plugin.name.in_(criteria["name"]))
         return query.order_by(models.TestGroup.priority.desc())
 
-    def PluginNametoCode(self, codes):
-        query = self.db.session.query(models.Plugin.code)
-        for count, name in enumerate(codes):
-            if "OWTF-" not in name:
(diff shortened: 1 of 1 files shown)


DePierre:
@delta24 Please be careful when merging PRs. OWTF is too unstable already :)

PR #705, opened 2016-06-06
[webui] React setup script

saganshul:
React setup instructions added to installation script.

## Reviewers
@delta24 @anantshri @tunnelshade @7a 

## How Has This Been Tested?
This is tested on ubuntu.

## Types of changes

- [ ] Other

### DIFF
--- a/install/install.py
+++ b/install/install.py
@@ -4,6 +4,7 @@ import os
 import sys
 import platform
 import argparse
+import subprocess
 from space_checker_utils import wget_wrapper
 
 import ConfigParser
--- a/install/setup_react.sh
+++ b/install/setup_react.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env sh
+#
+# This script install the node dependencies listed in package.json and build the bundle file using webpack.
+
+# bring in the color variables: `normal`, `info`, `warning`, `danger`, `reset`
+. "$(dirname "$(readlink -f "$0")")/utils.sh"
+
+# Instead of using apt-get to install npm we will nvm to install npm because apt-get installs older-version of node
+echo "${normal}[*] Installing npm using nvm.${reset}"
+wget https://raw.githubusercontent.com/creationix/nvm/v0.31.1/install.sh -O /tmp/install_nvm.sh
+bash /tmp/install_nvm.sh
+rm -rf /tmp/install_nvm.sh
+nvm install node
+nvm alias default node
+source ~/.bashrc
+echo "${normal}[*] npm successfully Installed.${reset}"
+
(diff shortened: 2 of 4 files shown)


install/install.py
```diff
@@ -110,6 +111,40 @@ def is_compatible():
             return True
 
 
+def setup_react_stack():
+    """Install the node dependencies listed in package.json and build the bundle file using webpack.
+
+    """
```
delta24: Docstring ends on the same line for a single-line comment.

delta24:
You should add this to the `distro_independent.cfg` file and it will called by the install script. Add an install script `setup_react.sh` which does the same thing as `setup_react_stack` and call it in the `distro_independent.cfg` like:

```
[React setup]
directory = /tmp/owtf-install/react_setup/%(Pid)s
command = sh %(RootDir)s/install/setup_react.sh
```

install/setup_react.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env sh
+#
+# This script install the node dependencies listed in package.json and build the bundle file using webpack.
+
+# bring in the color variables: `normal`, `info`, `warning`, `danger`, `reset`
+. "$(dirname "$(readlink -f "$0")")/utils.sh"
+
+sudo apt-get install curl
```
delta24: This should not be there.

saganshul: So can we assume that curl is installed?

delta24: Yes + should not assume that it is a Debian-based distro.

saganshul:
so you just want me to remove this line right?

install/setup_react.sh
```diff
@@ -0,0 +1,27 @@
...
+# This script install the node dependencies listed in package.json and build the bundle file using webpack.
+
+# bring in the color variables: `normal`, `info`, `warning`, `danger`, `reset`
+. "$(dirname "$(readlink -f "$0")")/utils.sh"
+
+# Instead of using apt-get to install npm we will nvm to install npm because apt-get installs older-version of node
+echo "${normal}[*] Installing npm using nvm.${reset}"
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.31.1/install.sh | bash
+nvm install node
+nvm alias default node
```
delta24: Don't use pipe to use on a script. This is a bad practice for an infosec tool :D

delta24:
Oops! my mistake :(
Change the react_setup directory in distro_independent.cfg to be `/tmp/owtf-install/react_setup` --> This is so that the installer installs it by default.

PR #738, opened 2016-09-14
[webui] React bundle build at the start of owtf

saganshul:
This will make owtf.py to build the bundle when owtf start.

## Reviewers
@delta24 @7a 

## Types of changes

- [x] Other

### DIFF
--- a/owtf.py
+++ b/owtf.py
@@ -193,6 +193,9 @@ def process_options(user_args):
         'Args': args
     }
 
+def build_react_bundle():
+    print("\033[93m[*] Building the react bundle. Plese wait..")
+    os.system("npm run build > /dev/null")
 
 def run_owtf(core, args):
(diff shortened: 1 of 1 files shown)


owtf.py
```diff
@@ -193,6 +193,9 @@ def process_options(user_args):
         'Args': args
     }
 
+def build_react_bundle():
+    print("\033[93m[*] Building the react bundle. Plese wait..")
+    os.system("npm run build > /dev/null")
```
delta24: Will this also run "webpack" to reload the config?

saganshul: It will just build the bundle.js in /includes/src/build. Internally it will run "$ NODE_ENV=production webpack". See package.json

delta24: Ok, great!

PR #755, opened 2016-10-09
Porting report to react

saganshul:
This ports the OWTF report to React.

## Reviewers
@delta24 @7a @tunnelshade 

## Types of changes

- [ ] Other


delta24:
@saganshul I don't there is any bug in the new report but I have not done the full code review. Have you tested this enough?

saganshul:
Yes! I have done enough testing for new report.

saganshul:
screenshot from 2016-10-25 12-13-55

includes/src/Report/Accordians.js
```diff
@@ -0,0 +1,157 @@
...
+                if (pluginDataList[i].user_rank > maxUserRank) {
+                    maxUserRank = pluginDataList[i].user_rank;
+                }
+            } else if (pluginDataList[i].owtf_rank != null && pluginDataList[i].owtf_rank != -1) {
+                if (pluginDataList[i].owtf_rank > maxOWTFRank) {
+                    maxOWTFRank = pluginDataList[i].owtf_rank;
+                }
+            }
+        }
+        testCaseMax = (maxUserRank > maxOWTFRank)
```
delta24: You could've written this conditional on the same line - more readability.

includes/src/Report/RankButtons.js
```diff
@@ -0,0 +1,92 @@
...
+
+class RankButtons extends React.Component {
+
+    render() {
+        var obj = this.props.obj;
+        var user_rank = obj['user_rank'];
+        var owtf_rank = obj['owtf_rank'];
+        if (user_rank in[0,
+            1,
+            2,
```
delta24: Should on the same line

includes/src/Report/Table.js
```diff
@@ -0,0 +1,171 @@
...
+                            <button ref={"editor_" + group + "_" + type + "_" + code} className="btn btn-default pull-right" onClick={this.handleEditor.bind(this, group, type, code)}>
+                                <i className="fa fa-pencil"></i>
+                                Notes
+                            </button>
+                        </th>
+                    </tr>
+                    <tr>
+                        <td colSpan="6">
+                            <textarea className="editor" style={{
+                                display: "none"
```
delta24: Same here.

framework/interface/templates/target.html
```diff
@@ -130,252 +25,62 @@ <h3 id="rankingTargetInfo"></h3>
...
-    $(".filterCheckbox").attr("checked", false);
-}
-
-function modifyFilter(checkbox, filter_type, data){
-    if (typeof mySpace.filter_data[filter_type] === 'undefined') {
-        mySpace.filter_data[filter_type] = [];
+    var mySpace = {
+        target_id: "{{ target_id}}",
+        worklist_api_url: "{{ worklist_api_url }}",
+        plugins_api_url: "{{ plugins_api_url }}?group=web&group=network",
```
saganshul: @delta24 Actually this is not hard-coded by me. I just copied it from plugin_report.html. I think no need to have this hard code thing now. I will remove it.

framework/db/poutput_manager.py
```diff
@@ -123,21 +123,34 @@ def GetAll(self, filter_data=None, target_id=None):
         return self.DeriveOutputDicts(results, target_id=target_id)
 
     @target_required
+    def GetAllPluginNames(self, filter_data=None, target_id=None):
```
delta24: Ok +1. Write comments (inline and header) so that it is easier to review and modify in the future.

includes/src/Report/Accordians.js
saganshul: @delta24  Actually I was using https://orktes.github.io/atom-react/ for auto indentation and stuff. This conditional multi-lined by this module. Just check once if you fine with this module.

framework/interface/templates/target.html
saganshul: No I am using auto indent from plugin.

includes/src/Report/Accordians.js
saganshul: https://github.com/felixge/node-style-guide#use-multi-line-ternary-operator

delta24:
Did you do any profiling tests using the Chrome dev tools to verify our assumptions that it will only load data incrementally - thereby increasing performance and less CPU/memory usage?

saganshul:
what do you mean by "load data incrementally"?

delta24:
The main advantage of using Rails is that you don't need to load (as an example) the whole plugin data every time there is an update (from a new plugin). This makes the API data grow to a very large size. React should then only update the new data, not rewrite the whole DOM.

Check the time it takes to load the whole report, and compare load times whenever a plugin finishes.

saganshul:
@delta24 ,
Pleas read this - http://stackoverflow.com/questions/35308928/how-to-avoid-re-rendering-the-whole-list-instead-of-adding-the-new-item-to-the-d

I have used key everywhere. React is able to do this using its virtual DOM.

Finally, 

As you suggested, I captured the DOM change when page load using https://chrome.google.com/webstore/detail/domlistener/jlfdgnlpibogjanomigieemaembjeolj

I saw that only changes is getting reflected in DOM. This means that ReactJS is somwhat is able to only update the new data, not rewrite the whole DOM.

Yes! I believe whenever a state changes ReactJS re-renders the whole DOM. But as you can read in above stake-overflow link if you assign a key, ReactJS is able to handle reflect only change.

Please correct me if I am wrong somewhere.

delta24:
Can you make the buttons more visible (boundary separation)?
image

saganshul:
@delta24 , PR updated.
ccc7526808286068c43e7815d746d59b

includes/src/Report/Accordians.js
```diff
@@ -38,6 +38,7 @@ class Accordians extends React.Component {
                     }
                     return (
                         <div className={(() => {
+                            console.log(key);
```
saganshul: shit! I forgot to remove that.

delta24:
There are 2 more bugs:

- The filters button renders in a weird way

selection_045

- The test groups should list only passive, semi-passive, grep, active, external.

saganshul:
@delta24 What do you mean by 2nd point?  As far as I know, this groups, type etc. are returned by backend and these are those group, type etc. which are present in your scan. If there is no plugin present for say "web" group "web" will not be present in Advance filter modal.

Please correct me If I am wrong somewhere.

delta24:
Each plugin group(except auxiliary)  - has only passive, semi-passive, grep, active, external plugin types.

saganshul:
@delta24 , I got it. It happened because we removed "?group=web&group=network" from plugin_api_url :P

delta24:
Yeah, think of something to make that change without hardcoding stuff.

saganshul:
@delta24 , so about the first bug, it is coming because sometime back-end is storing "" in status field that I don't know why. Well it should not be there.
But anyways for fix, I am just checking it again using JavaScript that coming string from back-end is NULL or not.

saganshul:
@delta24 , also on your https://github.com/owtf/owtf/pull/755#issuecomment-257014056 

* I don't understand why we are not showing auxiliary group.
* If we just want to have web and network, we have to hard-code it somewhere in backend or frontend. What can be other way, we can do it in back-end then other things will break. Because end-point is same /api/targets/<id>/poutput/<something_here> .Hence I think best way I could think is hard-code it here. Anyways I am not using myspace.plugins_api_url anywhere.

Let me know if you have some other way.

delta24:
@saganshul  I think the best solution would be to show those groups (for auxiliary) iff the user selects auxiliary group. Make it more user friendly. Something like:
Step1 : Choose the plugin group
Step 2: Choose plugin types (dynamically obviously - but store everything in JS (localStorage or object) and call/show when needed.

PR #763, opened 2016-10-10
[webui] Removes recently finished targets polling to improve preformance

saganshul:
Removes recently finished targets polling to improve performance

## Description
This is done because polling this endpoint is not much of importance and also it keeps server busy and hence decreases the performance.

## Reviewers
@delta24 @7a 

## Types of changes

- [ ] Other

### DIFF
--- a/framework/interface/templates/base.html
+++ b/framework/interface/templates/base.html
@@ -124,29 +124,6 @@
         $(selector).prop("disabled", true);
     }
 
-    // Function that fills the recently completed targets
-    function recentlyFinishedTargets() {
-        $.ajax({url: '/api/targets/recent/',
-                ifModified: true,
-                dataType:'json',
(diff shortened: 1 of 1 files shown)


delta24:
@saganshul if you removed the polling, then the feature is not useful anymore, right? Then it would be better to remove it.

saganshul:
@delta24 well I think it will still be useful because recently finished will be present on page load. Other than this what we can do we can increase the timeout period like 5 min or something?

delta24:
@saganshul IMO a better approach would be:
- remove the existing implementation for this
- Make a new notifications center and push all error messages, scan completion messages, etc there

This should use websockets for better performance and low overhead.

saganshul:
@delta24 that approach will be best. I think we should create issue regarding this as new feature and will work on it.

PR #776, opened 2016-11-14
Ignore nohup files

sachinkamath:
## Description

While running OWTF detached from the terminal using `nohup`, it generates a `.out` file. This PR adds this file to `.gitignore`

## How Has This Been Tested?

-NA-

-NA-

## Types of changes

- [x] Other


delta24:
@sachinkamath why are using nohup? Running OWTF as a daemon is a bad idea - you'd want to see the output (if there is!).

sachinkamath:
Ah great. I just wanted to demonize OWTF for using the web interface alone. It'd be really great if we can have an option to explicitly run OWTF as a daemon.

Issue #746, opened 2016-09-30
OWTF startup directory


saganshul:
Ideally, OWTF should never be run from its own directory because it does not make sense.
## Expected Behavior

On running OWTF from it own directory it should print "Please navigate to the working directory for your pentest and then run OWTF like this: /path/to/owtf.py"
This will create an owtf_review directory for that pentest/project"
## Current Behavior

It does not print this message and OWTF as usual.

delta24:
@saganshul @7a should we do something like this? 
1. Normal OWTF run should ask to create a new directory for a pentest (if not already in it).
2. For developers, we should add `--dev` to bypass this check
?

7a:
Even a developer should never run OWTF from its own directory:
We are talking output files being saved in a tool directory :)

I think the best way to avoid this mess is simply to do something like this:
Option 1) Refuse to run OWTF if on the same directory.
or
Option 2) Require an output directory parameter when OWTF is run

Other thoughts in https://github.com/owtf/owtf/issues/766 :)

DarKnight24:
@delta24 Should we refuse to run if on the same directory?

delta24:
@DarKnight24 yes.

delta24:
Fixed in https://github.com/owtf/owtf/pull/789.

PR #779, opened 2016-12-05
Fixes #766. Owtf now refuses to run in its own directory

### LINKED ISSUE #766: OWTF installer recommendation misleads new users
OWTF should never, ever, be run from its own directory, this makes absolutely no sense since that will result in creating the output files for a given pentest within the OWTF directory.

In the real world, an OWTF user should run OWTF *always* like:

cd /path/to/pentest/directory
/path/to/owtf.py

This results in the output files being saved in the correct directory for a given pentest, instead of the OWTF directory.

Unfortunately, the installer is recommending users to do exactly this, which has led to confusion in the past, this should be replaced with an explanation like above:

[*] Start OWTF by running './owtf.py' in parent directory

We had even distro maintainers trying to call OWTF like this, maybe we should even go a step further and literally refuse to run OWTF when it is on its own directory.

Thoughts? :)

ujjwal96:
## Description

Whenever owtf is run from its own directory it exits giving an error message.
Also the message after installation is not misleading anymore.

## Related Issue

#766 #746

### DIFF
--- a/owtf.py
+++ b/owtf.py
@@ -255,4 +255,9 @@ def main(args):
 
 
 if __name__ == "__main__":
+
+    if 'owtf.py' in os.listdir(os.getcwd()) or 'owtf' == os.getcwd()[:-4:]:
+        usage("Start owtf in a different directory. 'cd /path/to/pentest/directory' \n'/path/to/owtf.py'")
+        exit(-1)
+
     main(sys.argv)
--- a/install/install.py
+++ b/install/install.py
@@ -116,7 +116,7 @@ def finish(error_code):
             Colorizer.normal("[*] Visit https://github.com/owtf/owtf for help ")
         else:
             Colorizer.success("[*] Finished!")
-            Colorizer.info("[*] Start OWTF by running './owtf.py' in parent directory")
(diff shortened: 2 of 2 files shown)


ujjwal96:
@delta24 Can you review this and tell me if there's anything else that needs to be done?

DarKnight24:
@ujjwal96 This is not the correct fix for #766.

ujjwal96:
@DarKnight24 What needs to be done?
The description said that owtf should refuse to run when in its own directory. I also changed the message that comes after the installation.

DarKnight24:
@ujjwal96 have a look at issue #746

owtf.py
```diff
@@ -255,4 +255,9 @@ def main(args):
 
 
 if __name__ == "__main__":
+
+    if 'owtf.py' in os.listdir(os.getcwd()) or 'owtf' == os.getcwd()[:-4:]:
```
DarKnight24: What is the use of second condition?

delta24:
Closing in favor of a duplicate, recent PR.

ujjwal96:
@delta24 Would you mind telling me how the recent PR was a better way to do it? I'd like to keep those things in mind for my future PRs, if any :)

PR #789, opened 2016-12-23
Fixes #766. OWTF will never run within its own directory.

### LINKED ISSUE #766: OWTF installer recommendation misleads new users

affinity7:
The OWTF will exit if executed from its directory

## Description

Added a check confirming the OWTF is not executed from its directory.

## Related Issue

#766  #746 

## Reviewers

@delta24 @DarKnight24 @7a 

## How Has This Been Tested?

Tested it by running it from different directories.

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/owtf.py
+++ b/owtf.py
@@ -255,4 +255,9 @@ def main(args):
 
 
 if __name__ == "__main__":
+
+    if os.path.dirname(os.path.abspath(sys.argv[0])) == os.getcwd():
+        usage("Start owtf in a different directory. 'cd path/to/pentest/directory; ./path/to/owtf.py'")
+        exit(-1)
+
     main(sys.argv)
--- a/owtf.py
+++ b/owtf.py
@@ -255,9 +255,7 @@ def main(args):
 
 
 if __name__ == "__main__":
-
-    if os.path.dirname(os.path.abspath(sys.argv[0])) == os.getcwd():
+    if os.getcwd().find(os.path.dirname(os.path.abspath(sys.argv[0])))==0:
         usage("Start owtf in a different directory. 'cd path/to/pentest/directory; ./path/to/owtf.py'")
         exit(-1)
-
     main(sys.argv)
--- a/owtf.py
+++ b/owtf.py
@@ -255,7 +255,7 @@ def main(args):
 
 
 if __name__ == "__main__":
-    if os.getcwd().find(os.path.dirname(os.path.abspath(sys.argv[0])))==0:
+    if os.getcwd().find(os.path.dirname(os.path.abspath(sys.argv[0]))) == 0:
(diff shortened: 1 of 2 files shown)


owtf.py
```diff
@@ -255,4 +255,9 @@ def main(args):
 
 
 if __name__ == "__main__":
+
```
DarKnight24: Do not add line after the function name. Remove this extra line.

owtf.py
```diff
@@ -255,4 +255,9 @@ def main(args):
...
+    if os.path.dirname(os.path.abspath(sys.argv[0])) == os.getcwd():
```
DarKnight24: Instead of this check, check if OWTF is ran from its directory/sub-directories. Check for the substring, @delta24 what is your view?

delta24: @DarKnight24 what substring?

DarKnight24: If executed file folder is substring of current working directory, it is so then we can say that OWTF is ran from its directory or subdirectory.

delta24: ok, that makes sense.

owtf.py
```diff
@@ -255,9 +255,7 @@ def main(args):
 
 
 if __name__ == "__main__":
-
-    if os.path.dirname(os.path.abspath(sys.argv[0])) == os.getcwd():
+    if os.getcwd().find(os.path.dirname(os.path.abspath(sys.argv[0])))==0:
```
delta24: Spaces around the comparison missing.

owtf.py
```diff
@@ -255,4 +255,7 @@ def main(args):
 
 
 if __name__ == "__main__":
+    if os.getcwd().find(os.path.dirname(os.path.abspath(sys.argv[0])))==0:
+        usage("Start owtf in a different directory. 'cd path/to/pentest/directory; ./path/to/owtf.py'")
+        exit(-1)
```
delta24: Use `sys.exit` here. (exit is a helper for the interactive shell - sys.exit is intended for use in programs.)

delta24:
@affinity7 Thank you for the PR :)!

PR #793, opened 2017-01-03
React Documentation

saganshul:
PR submits the functional documentation for React Code written(New UI)

## Reviewers
@delta24 @7a @DarKnight24 

## Types of changes

- [x] Other

### DIFF
--- a/includes/src/Dashboard/Chart.js
+++ b/includes/src/Dashboard/Chart.js
@@ -1,6 +1,12 @@
 import React from 'react';
 import {Pie} from 'react-chartjs';
 
+/**
+ *  React Component to create one entry of chart legend.
+ *  It is child components which is used by ChartLegend Component
+ *  Receives {"color": <color hex>, "id": <id>, "value": <int>, "label": <severity>} as an JS object where color represent color of that severity.
+ */
+
 class Severity extends React.Component {
 
     render() {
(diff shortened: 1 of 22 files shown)


delta24:
@saganshul the docs look complete. Before I merge, have you used React without the JSX syntax?

saganshul:
@delta24, No, I am using JSX. You can see my every render function, it is JSX.

delta24:
Then the file extensions should be ideally `.jsx`, right?

saganshul:
Well It doesn't mater actually. But I can explain why I use .js instead.

We're going to have less trouble if we use .js for both JS and JSX. Also, to me it makes sense because we're processing both with the same Babel configuration anyway (or we could parse JSX files with a separate configuration, though I'm not sure if that's worth it).
With .jsx we'd either have to remember to put the extension when importing modules:
> import foo from './foo.jsx';

or we'd have to configure our loader to look for .jsx extension as well. Also, sometimes we are not really sure whether we're going to write some JSX in a file, so we might use the .js extension at first, then changing it to .jsx later would be an unnecessary "move" change in version control. Another example, if we tests have both .js and .jsx files, we'd probably have to put extra effort to run all files, not just .js.

delta24:
@saganshul This [thread](https://github.com/facebook/react/issues/3582#issuecomment-89411479) sums up why using `.js` extension for JSX code is bad.

Can you convert it to have `.jsx` everywhere?

PR #800, opened 2017-01-08
Fixes #790. Moved default config files to configuration folders and checks for user profiles.

### LINKED ISSUE #790: [config] Move the framework configuration file out of framework directory
The `framework_config.cfg` file is currently inside the `framework/config` directory which is bad.
Instead there should be a default configuration file which is there and the user modified configuration file is not committed.

 The `-m` profiles option provides this functionality for the loading user profiles but not the framework configuration.

#### Solution
- [ ] Move the `framework_config.cfg` outside the framework directory.
- [ ] Rename it as `framework_config.cfg.default`.
- [ ] At run-time, check for the file `framework_config.cfg`. If it does not exists, create a copy of the default configuration file for usage, otherwise load the user modified `framework_config.cfg` file.

or maybe even move all configuration files to `~/.owtf`? @DarKnight24 @7a?

affinity7:
## Description

Moved the configuration files into `$ROOTDIR_owtf/configuration` folder and checks for user profiles in `~/.owtf/configuration` folder incorporated. If present, it loads those files and if absent, it creates a copy of the default files in that folder.

## Related Issue

#790 

## Reviewers

@delta24 @DarKnight24 @DePierre 

## How Has This Been Tested?

By running OWTF.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Other

### Checklist:


delta24:
@affinity7 You missed something - Travis check is failing because it cannot find a config file. Please take a look. https://travis-ci.org/owtf/owtf/builds/190048252

configuration/frameworks/framework_config.cfg.default
```diff
@@ -45,11 +45,11 @@ TARGETS_DIR: targets
```
DarKnight24: Yes, all the profiles should be in configuration folder IMO. It makes no sense as only one config files will be present for each case.

framework/config/config.py
```diff
@@ -57,7 +58,18 @@ def __init__(self, root_dir, owtf_pid):
...
+            os.makedirs(path)
+        except OSError:
+            if not os.path.isdir(path):
+                raise
+        path_config = os.path.join(path, 'framework_config.cfg')
+        if not os.path.isfile(path_config):
+            with open(path_config,"w+") as f:
+                f.close()
+            shutil.copy(os.path.join(self.RootDir, 'configuration', 'framework_config.cfg.default'), path_config)
+        self.LoadFrameworkConfigFromFile(os.path.join(path_config))
```
DarKnight24: This `os.path.join` is redundant

framework/db/config_manager.py
```diff
@@ -29,6 +30,18 @@ def IsConvertable(self, value, conv):
 
     def LoadConfigDBFromFile(self, file_path):
         # TODO: Implementy user override mechanism
+        path = os.path.join(os.path.expanduser("~"), os.path.join('.owtf', 'configuration'))
```
DarKnight24: Do not take absolute path in the codebase. Use `framework_config.cfg` to take appropriate value.
What is your view @delta24

DarKnight24: Add a variable for the configuration files with value `~/.owtf/configuration` to avoid path joins

DePierre:
@affinity7 thanks for your PR. As pointed out, there are a couple of comments you should address before we can merge it.

I have a lot of comments behind the logic of your redundant code your wrote for each configuration file.

```python
# This should be a framework function to get the OWTF 'configuration' directory
path = os.path.join(os.path.expanduser("~"), os.path.join('.owtf', 'configuration'))
try:
    # This should be part of the installation script
    os.makedirs(path)
except OSError:
    # What's the logic behind? Why would you silently catch unexpected exceptions?
    if not os.path.isdir(path):
        raise
# The filename should not be hardcoded (although it was before but let's improve the code base please)
path_mapping = os.path.join(path, 'mapping_manager.cfg')
if not os.path.isfile(path_mapping):
    # What's that? A hack to 'touch' the file? It looks really ugly
    with open(path_mapping,"w+") as f:
        f.close()
    # OK so this is supposed to implement that "if no user specified, use the default one"
    # It seems that the logic is correct in `config.py` but it doesn't seem correct here.
    # I mean, does it even work? `file_path` doesn't seem to be defined at that point.
    shutil.copy(file_path, path_mapping)
file_path = path_mapping
```

Then please consider heavily refactoring the code and also investigate the Travis issue https://travis-ci.org/owtf/owtf/builds/190048252 as pointed out by @delta24 

Also, this is a breaking change and should be tagged as such. We are breaking the previous OWTF behavior in a sense that if I, as a user, previously changed the configuration of OWTF, my changes will be lost due to this patch.

framework/config/config.py
```diff
@@ -57,7 +58,18 @@ def __init__(self, root_dir, owtf_pid):
...
+        try:
+            os.makedirs(path)
+        except OSError:
+            if not os.path.isdir(path):
+                raise
+        path_config = os.path.join(path, 'framework_config.cfg')
+        if not os.path.isfile(path_config):
+            with open(path_config,"w+") as f:
+                f.close()
+            shutil.copy(os.path.join(self.RootDir, 'configuration', 'framework_config.cfg.default'), path_config)
```
DarKnight24: Use `shutil.copyfile` instead. You will not require the above command to create a file beforehand.

delta24:
@affinity7  I think a better approach would be to copy the default configuration directory to `~/.owtf` during OWTF installation and just update the paths in the config files. @DePierre @DarKnight24 ?

DePierre:
@delta24 That's pretty much what I said in my comments. I agree that creating configuration directory and copying the default configuration files should be part of the installation script.

@affinity7 basically, the logic would be:
```python
# Installation phase
create_owtf_directory('~/.owtf')
copy_default_dotfiles()  # copy general.cfg under ~/.owtf/configuration/default/, framework.cfg, etc.
```

Then in the OWTF python code, when starting the framework, you would follow some logic like this:
```python
# OWTF python starting code
load_user_conf_or_default(CONF_TYPE):  # e.g. GENERAL, FRAMEWORK, etc.
    if user_conf does not exist:
        return load_default(CONF_TYPE)
    return load_user_conf(CONF_TYPE)
```

DePierre:
@affinity7 Still feeling motivated to continue working on the PR?

affinity7:
@DePierre yes I am currently working on it. I will modify the pr by today.

DePierre:
@affinity7 Awesome! Just checking ;)

I know that sometimes we end up asking much more than initially thought. But I believe that this PR is a step towards getting rid of our awful installation process.

affinity7:
@DePierre Do we need to create two separate folders- `default` and `user` in `/.owtf/configuration/` folder?

delta24:
@affinity7 no. The preference (when loading the configuration) should be given to the files in `$HOME/.owtf/configuration` before `$OWTF_DIR/configuration`.

framework/config/config.py
```diff
@@ -84,6 +72,13 @@ def initialize_attributes(self):
         for type in CONFIG_TYPES:
             self.Config[type] = {}
 
+    def SelectConfigFilePath(self, file_path):
```
delta24: Make the function names to use the snakecase - `select_config_path`. The other function names in the file will be change in a short while.

install/install.py
```diff
@@ -264,4 +265,11 @@ def danger(cls, string):
     Colorizer.info("[*] Last commit hash: %s" % owtf_last_commit())
     check_sudo()
     installer_status_code = install(sys.argv[1:])
+
+    # Copying config files
+    path = os.path.join(os.path.expanduser('~'), '.owtf', 'configuration')
+    create_directory(path)
+    path_config = os.path.join(root_dir, 'configuration')
+    dir_util.copy_tree(path_config, path)
```
delta24: Does this throw error if some of the files in the destination exist?

affinity7: @delta24 no. I checked by deleting some files. It copies the remaining files without any error.

framework/config/config.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, root_dir, owtf_pid):
         # Available profiles = g -> General configuration, n -> Network plugin
         # order, w -> Web plugin order, r -> Resources file
         self.initialize_attributes()
-        self.LoadFrameworkConfigFromFile(os.path.join(self.RootDir, 'framework', 'config', 'framework_config.cfg'))
+        self.LoadFrameworkConfigFromFile(self.select_config_path(os.path.join(os.path.expanduser('~'), '.owtf', 'configuration', 'framework_config.cfg')))
```
DePierre: Shouldn't
``self.select_config_path(os.path.join(os.path.expanduser('~'), '.owtf', 'configuration', 'framework_config.cfg')``
be in fact
``self.select_config_path('framework_config.cfg')``
and the `os.path.thingy` moved to `self.select_config_path()`?

configuration/framework_config.cfg
```diff
@@ -45,11 +45,11 @@ TARGETS_DIR: targets
 
 
 # ---------------------- Default profile settings ---------------------- #
-DEFAULT_GENERAL_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/general/default.cfg
-DEFAULT_RESOURCES_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/resources/default.cfg
+DEFAULT_GENERAL_PROFILE: ~/.owtf/configuration/general.cfg
+DEFAULT_RESOURCES_PROFILE: ~/.owtf/configuration/resources.cfg
+DEFAULT_MAPPING_PROFILE: ~/.owtf/configuration/mappings.cfg
 DEFAULT_WEB_PLUGIN_ORDER_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/plugin_web/order.cfg
 DEFAULT_NET_PLUGIN_ORDER_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/plugin_net/order.cfg
```
DePierre: Shouldn't profiles configuration files be part of the configuration directory?

configuration/framework_config.cfg
```diff
@@ -45,11 +45,11 @@ TARGETS_DIR: targets
 
 
 # ---------------------- Default profile settings ---------------------- #
-DEFAULT_GENERAL_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/general/default.cfg
-DEFAULT_RESOURCES_PROFILE: @@@FRAMEWORK_DIR@@@/profiles/resources/default.cfg
+DEFAULT_GENERAL_PROFILE: ~/.owtf/configuration/general.cfg
+DEFAULT_RESOURCES_PROFILE: ~/.owtf/configuration/resources.cfg
+DEFAULT_MAPPING_PROFILE: ~/.owtf/configuration/mappings.cfg
```
DePierre: Silly question but in the case we want to support Windows, wouldn't `~/` fail? Can't we use the `os` module to expand the user home directory for us?

framework/config/config.py
```diff
@@ -72,6 +72,13 @@ def initialize_attributes(self):
         for type in CONFIG_TYPES:
             self.Config[type] = {}
 
+    def select_config_path(self, file_path):
```
DePierre: The name should be more explicit, such as `user_path_or_default`. Also there should be a docstring to explain what is does, what it returns, etc.

framework/config/config.py
DePierre: What I mean is why is config.py the only one that has such a ugly looking parameter compared to config_manager.py for isntance

DarKnight24: @DePierre Framework config is the only config file on which owtf completely depends. This file contains important parameters and location of other config files. So the path of this file has to be hard coded and fixed.

DePierre: @DarKnight24 it makes sense, nvm then :)

affinity7:
@delta24 @DePierre  Could you please tell me if I need to add anything else to this issue?

DePierre:
@affinity7 Did you have a look at my comments? I asked a couple of questions regarding the PR and I would have appreciated some answers if possible. I am not very familiar with this part of OWTF so my questions might be dumb or nonsensical. If so, please let me know why ;)

configuration/framework_config.cfg
affinity7: @DePierre So should the plugin files also be moved to the `configuration` directory?

DePierre: @affinity7 I would believe that they are configuration files as well so it would seem logical to move them into the configuration folder. @delta24 any thought?

DePierre:
@affinity7 Would it be possible to rebase your PR on the current develop head? It would be nice to have travis working again for your PR so we can check whether something broke ;)

affinity7:
@delta24 @DePierre @DarKnight24 Can anyone explain why Travis is not getting the `db.cfg` file?

DePierre:
@affinity7 I believe that the reason Travis fails is because you missed some references in your changes.

For instance, the DB scripts to start/configure the database is still trying to use the old paths (see https://github.com/affinity7/owtf/blob/fix-790/scripts/db_setup.sh#L68). You can see the error during the installation of OWTF in Travis:

https://travis-ci.org/owtf/owtf/builds/196123998#L1469
```
[*] Running following command
sh /home/travis/build/owtf/owtf/install/db_config_setup.sh /home/travis/build/owtf/owtf
grep: /home/travis/build/owtf/owtf/framework/config/framework_config.cfg: No such file or directory
```

I would highly recommend testing your changes on a clean OWTF install each time. A possible setup would be one machine where you develop, and one VM with a snapshot just before installing owtf. You develop on the first machine and when you think it is ready to be pushed, you perform a clean install in the VM and see if nothing broke.

In your case, when running OWTF everything works fine because the DB is initialized once during install.

affinity7:
@DePierre Thanks. Will work on it.

DePierre:
@affinity7 It seems that with your latest commits, the installation is successful :) The error you see is related to another commit (see https://github.com/owtf/owtf/issues/807#issuecomment-275830380) so we'll have to fix that first before merging.

In the meantime, could you look at https://github.com/owtf/owtf/pull/800/files#diff-e84a293c44758d5fff68428ace8d0159R51 and see if it would make sense to move them to `/configuration` as well? Since `DEFAULT_GENERAL_PROFILE`, which was in `/profiles`, has been moved to `/configuration`, it would make sense to move `DEFAULT_WEB_PLUGIN_ORDER_PROFILE` and `DEFAULT_NET_PLUGIN_ORDER_PROFILE` as well.

delta24:
Your changes with some modification have been merged in https://github.com/owtf/owtf/commit/6c1ff855773914919f9c3dc3d213bdf0dce16ab4. thank you @affinity7 !

DePierre:
@delta24 We should have waited for Travis to be fixed before merge the PR IMO.

delta24:
@DePierre yes. My bad.

PR #803, opened 2017-01-12
Corrected if condition in db_run script

DarKnight24:
## Reviewers

@delta24


scripts/db_setup.sh
```diff
@@ -44,11 +44,9 @@ postgresql_fix() {
       if [ "$service_bin" != "1" ]; then
         sudo service postgresql restart
         sudo service postgresql status | grep -q "Active: active"
-        status_exitcode="$?"
```
delta24: Why remove this?

DarKnight24: It is not used anywhere

PR #804, opened 2017-01-17
Changes .js to .jsx

saganshul:
Taking care of this [thread](https://github.com/facebook/react/issues/3582#issuecomment-89411479) all .js files renamed to jsx and webpack is configured accordingly.

## Reviewers
@delta24 @7a 

## Types of changes

- [x] Other


includes/src/Report/Table.js
```diff
@@ -127,7 +127,7 @@ class Table extends React.PureComponent {
                             if (output_path !== undefined) {
                                 return (
                                     <td>
-                                        <a href={"/output_files/" + output_path} target="_blank" className="btn btn-primary">Browse</a>
+                                        <a href={"/output_files/" + output_path} target="_blank" className={output_path === null ? "btn btn-primary disabled":"btn btn-primary"}>Browse</a>
```
DePierre: Would it be smarter to have something like this? (jinja syntax but you get the idea)

```django
<a className="btn btn-primary{% if output_path is None %} disabled{% endif %}"></a>
```

Issue #807, opened 2017-01-26
Travis is failing again


DePierre:
Since 46f305b2e3ec85849b149053d715a60ee6509ac3, it appears that Travis is failing again (see https://travis-ci.org/owtf/owtf/builds/195367688).

This time with the following traceback during `pip` installation:

```py3tb
Collecting pyparsing (from packaging>=16.8->setuptools->PyHamcrest==1.9.0->-r ./owtf_testing/owtf_testing.pip (line 1))
  Using cached pyparsing-2.1.10-py2.py3-none-any.whl
Installing collected packages: six, appdirs, pyparsing, packaging, setuptools, PyHamcrest, backports.ssl-match-hostname, singledispatch, certifi, backports-abc, tornado, funcsigs, pbr, mock
  Found existing installation: six 1.10.0
    Uninstalling six-1.10.0:
      Successfully uninstalled six-1.10.0
  Rolling back uninstall of six
Exception:
Traceback (most recent call last):
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/wheel.py", line 247, in move_wheel_files
    prefix=prefix,
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/pip/locations.py", line 140, in distutils_scheme
    d = Distribution(dist_args)
  File "/home/travis/.virtualenvs/owtf/lib/python2.7/site-packages/setuptools/dist.py", line 320, in __init__
    _Distribution.__init__(self, attrs)
  File "/opt/python/2.7.12/lib/python2.7/distutils/dist.py", line 287, in __init__
    self.finalize_options()
[...]

PR #818, opened 2017-02-13
Revert "Added sudo access and corrected update command"

DePierre:
Test travis build (see #807)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,9 +23,9 @@ install:
     - pip install --upgrade -r ./owtf_testing/owtf_testing.pip
 
 before_script:
-    - cp -f ./owtf_testing/profiles/plugin_web/groups.cfg ./profiles/plugin_web/groups.cfg
-    - cp -f ./owtf_testing/profiles/plugin_net/groups.cfg ./profiles/plugin_net/groups.cfg
-    - cp -f ./owtf_testing/profiles/plugin_aux/groups.cfg ./profiles/plugin_aux/groups.cfg
+    - cp -f ./owtf_testing/profiles/plugin_web/groups.cfg ~/.owtf/configuration/profiles/plugin_web/groups.cfg
+    - cp -f ./owtf_testing/profiles/plugin_net/groups.cfg ~/.owtf/configuration/profiles/plugin_net/groups.cfg
+    - cp -f ./owtf_testing/profiles/plugin_aux/groups.cfg ~/.owtf/configuration/profiles/plugin_aux/groups.cfg
     - git config --global user.email "tasty@mac.test"
     - git config --global user.name "Tasty Test"
     - git am < ./owtf_testing/patch/0001-unittest-Patch-to-reuse-proxy-address.patch
--- a/install/debian/install.sh
+++ b/install/debian/install.sh
@@ -45,9 +45,9 @@ sudo sh -c "echo 'deb http://repo.kali.org/kali kali-rolling main contrib non-fr
 
 # Patch script for debian apt
 echo "${normal}[*] Adding apt preferences in order to keep Debian free from Kali garbage as much as possible :P${reset}"
-sudo sh "$RootDir/install/debian/pref.sh"
+sh "$RootDir/install/debian/pref.sh"
 
-sudo apt-get update
+sudo "$apt_wrapper_path" update
 
 echo "${normal}[*] Installing missing tools${reset}"
 sudo -E "$apt_wrapper_path" install lbd arachni tlssled nmap nikto skipfish w3af-console dirbuster wapiti hydra waffit \


DePierre:
@delta24 @DarKnight24 so reverting ad02c7bfd9b27f372aa13e6d95e93932496a85f9 leads to two interesting results:

+ Travis build takes almost 40 minutes
+ Travis build never ends

I will see how it goes when reverting https://github.com/owtf/owtf/pull/800.

By the way, I am not sure to understand why you break Travis, continue pushing untested commits and merging untested PR, and then just move along, hoping that I would fix it. OWTF has had many stability issues and I thought that it was part of OWTF's priority to become more stable now.

Isn't it a priority anymore?

DePierre:
I am aware that the test suites have bugs and issues due to OWTF not properly releasing resources and because I made mistakes that I haven't found and fixed yet. And I will be happy to fix those issues if one can point them to me.

However, breaking other's people commits because they don't care about breaking the test suites is another thing.

DePierre:
@delta24 @DarKnight24 I fixed the travis configuration to copy the user-specified configuration to the right place (I copied them to their default place but then realized that 6c1ff855773914919f9c3dc3d213bdf0dce16ab4 was copying the default configuration to the user's configuration file).

@delta24 6c1ff855773914919f9c3dc3d213bdf0dce16ab4 is not correct IMO. By default, the configuration files should be loaded from the user's owtf directory if they exist. If not, then OWTF should fallback on the default configuration. With 6c1ff855773914919f9c3dc3d213bdf0dce16ab4, OWTF will assign the default configuration to the user's owtf directory despite the fact that the user didn't ask for it. What do you think? Maybe I am misunderstanding the logic.

delta24:
@DePierre When the user first installs OWTF, the default configuration is copied to `~/.owtf` directory. Then if the user wishes to make any configuration changes,  the user can directly do that without messing up the parent OWTF directory. This serves 2 purposes:
- We keep the repository clean of configuration files
- If in the future we want to package OWTF for different distros/package managers, it becomes a lot easier if the configuration files live outside the repository (by default).

DePierre:
@delta24 Fair point. Based on https://github.com/owtf/owtf/pull/800, I remembered that the default configuration files would not be copied by default but re-reading the thread, I realized that I remembered wrong.

DarKnight24:
@delta24 @Depierre IMO it makes no sense to keep wrong code in the codebase. I think we should comment those command instead of reverting the commit.

delta24:
@DarKnight24 you mean remove instead of commenting, right?

DarKnight24:
@delta24 @DePierre We should remove the code add proper comment to fix the problem of installing package from kali source in future.

The debian script is currently not installing any package which are absent in debian repo.
One of the problem is travis is running on Ubuntu 12.04 and packages from kali requires newer version which are not conflicting with 12.04 version. I recommend to update it to 16.04 using docker. This will help me in correcting the debian script to install all the required packages.
IMO we should also add kali to travis testing.

delta24:
@DarKnight24 Ok, got your point. Btw Travis does not support Kali Linux (only Debian, Ubuntu). Look at the commits in this PR - atm Travis build is passing. 

@DePierre should we just remove the code responsible for this and add a #TODO to fix it if a clean solution is possible?

DePierre:
@delta24 To be honest no idea since I still don't understand what it is supposed to be doing. You and @DarKnight24 are better suited to know what we should do.

+ If Travis fails because it's a problem with the the test suites, I'll be happy to troubleshoot the issue with you and attempt to fix it.
+ If Travis fails because it doesn't support what we are trying to do, then we're screwed. But we could maybe update the test suites to use another tool than skipfish for instance.
+ If Travis fails because the commit broke something, then the commit should first and foremost reverted. Then a PR should be created to discuss why the commit is breaking, if it's expected, how we can fix it, etc.

You want me to rebase the PR on what? It was based on the develop HEAD when I created it so I am not sure to understand what you mean. Basically all I did was revert the commit that was messing with Skipfish installation and updated .travis configuration file to copy the lightweight OWTF configuration files to their new locations (lightweight in a sense that only a subset of plugins will be run, not all of them, to limit the time required to run all the tests).

Let me know.

delta24:
@DePierre the rebase thing was for @DarKnight24 's PR for fixing the said install script.

DarKnight24:
@delta24 @DePierre I have tried to install owtf on Ubuntu 16.04 (Look here) and found that install script runs successfully and installs kali packages too :). 
I had tried to run test on which travis was failing previously also ran successfully.

I think we should configure travis to run Ubuntu 16.04 using docker.

delta24:
@DarKnight24 I don't think we can configure how Travis runs - they actually run the builds on a Ubuntu Docker container.

delta24:
@DePierre should I merge this? I need to merge @DarKnight24 's changes for the install script but I'd prefer if the Travis is actually working first and see what it tells us.

DePierre:
@delta24 I dont know :s After reading through the thread again, I am not sure to understand what's up.

@DarKnight24 If I understood correctly, your commit is good and is working as expected on Kali. But it doesn't work on Ubuntu.
And if I understood correctly, we cannot modify Travis to use something else than Ubuntu.
Is that correct?

So should I update the tests to not use Skipfish? Or should we try to find another solution?

delta24:
@DePierre I mean that we merge this commit as it restores Travis to a working condition. Then I can rerun Travis on @DarKnight24 's PR to make sure it does not break again.

DePierre:
@delta24 sure! I think that it would be wise to have Travis working again so that commits and PRs are at least tested once.

Then we can take time to continue discussing this issue and find the right solution.

PR #826, opened 2017-02-17
Bug in ranking a plugin

saganshul:
## Description
Previously, if user ranks a plugins an api call only updates **user_rank** to corresponding rank.
Ideally we should also update owtf_rank because for suppose before rank PATCH call owtf ranks plugin to for say "info" but if we update the plugin user_rank to "passing" (because sometimes automated tools flag many things as findings which are really stupid) max_owtf_rank will still remain higher than max_user_rank for that target which is a bug actually. 
For example:
if only partially ranked, let's say, 3 automated plugins have an automated ranking and the human ONLY overwrote ONE, then **you still have to use the max of (automated, human) until the human overwrites the other automatically ranked plugins**.

Point is **if user_rank exist for some plugin then there is no requirement of owtf_rank.**

The point of **owtf_rank** is to attract the attention of the penetration tester towards the seemingly weaker hosts.

## Reviewers
@delta24 @7a 

## Types of changes

- [ ] Other

### DIFF
--- a/framework/db/poutput_manager.py
+++ b/framework/db/poutput_manager.py
@@ -172,6 +172,7 @@ class POutputDB(BaseComponent, PluginOutputInterface):
                     if isinstance(patch_data["user_rank"], list):
                         patch_data["user_rank"] = patch_data["user_rank"][0]
                     obj.user_rank = int(patch_data["user_rank"])
+                    obj.owtf_rank = -1
                 if patch_data.get("user_notes", None):
                     if isinstance(patch_data["user_notes"], list):
                         patch_data["user_notes"] = patch_data["user_notes"][0]


delta24:
@saganshul what happens if I revert the user_rank on the UI? It really should revert back to owtf_rank.

saganshul:
@delta24 I am not getting what do mean by revert. Is there any option for that in UI? I can only see "passing", "info", "low", "medium", "high", "critical".

DePierre:
@delta24 @saganshul It's been a while since I added the automated ranking so please forgive me if it's not 100% accurate but IIRC, once you override a plugin's ranking, the automated ranking is lost forever.

The justification behind that is that the automated ranking is supposed to provide a first insight for the user to help him focus on what seems to be the highest priority. Then if the user overrides the ranking, it means that he knew better and therefore the automated ranking became obsolete.

@saganshul If I understood your description of the bug correctly then yes, the user ranking has the highest priority. If plugin A was automatically ranked high-risk but the user changed it to passing, then passing should be kept.

But this behavior was implemented long time ago. Do we know what commit introduced the bug? Since when does it behave like this?

DePierre:
@saganshul Could you rebase your PR to include the updated travis configuration?

saganshul:
@DePierre Rebased and is in sync with develop.

viyatb:
@saganshul ah, ok.
@DePierre since the automatic ranking is lost after a user manually assigns the ranking, then the behavior implemented in this commit is correct. Thoughts?

DePierre:
@viyatb I am not sure. I asked since when this bug exist but I didn't receive any answer yet so I would appreciate if @saganshul could answer.

I'll have to look closer.

saganshul:
@DePierre @viyatb well I don't know when this bug is introduced. @7a pointed it out during project testing.

DePierre:
Anyone willing to `git bisect`?

viyatb:
@DePierre if you see the "Blame" view on https://github.com/owtf/owtf/blame/develop/framework/db/poutput_manager.py, then it looks like it always this behavior.

DePierre:
@viyatb This is where I am not sure that this is the case.

I am not saying that there was no bug but I explicitly remember testing for this behavior when integrating PTP to OWTF. This is why I am surprised and asking if someone could bisect, because it seems that something somewhere else changed. The change either introduced a bug, or uncovered this one and I would like to understand which is which.

Another possibility is that I remember it wrong and that removing user's ranking would restore the automatic ranking.

viyatb:
@DePierre I think that the behavior was the same from the beginning.

DePierre:
@viyatb You mean that removing the user ranking would restore the automated ranking? It's been almost 3 years so I have to be honest, I am not certain to remember.

viyatb:
@DePierre Not restore. Currently, the user cannot revert his/her ranking - once they modify any ranking, they have to supply a manual ranking. This PR conforms to that philosophy.

DarKnight24:
@viyatb @DePierre @saganshul I didn't understand the use of removing owtf_ranking. IMO client (here webui) implementation must be written such that user_ranking is given priority.

DePierre:
@DarKnight24 That was the initial idea yes. owtf_ranking attracts the eyes of the user and then the user overrides it with user_ranking. Then only user_ranking should matter since the user knows best.

PR #828, opened 2017-02-19
Fixed issue in path of plugin groups

DarKnight24:
This bug was introduced in [this](https://github.com/owtf/owtf/commit/6c1ff855773914919f9c3dc3d213bdf0dce16ab4) commit.

### Current Behaviour:
If any configuration file is missing in `~/.owtf/` folder then current owtf expects it to be strictly in `@@@RootDir@@@/configuration/`. As the default plugin config files are in `@@@RootDir@@@/configuration/profiles/`, this makes default files unavailable.

### Expected Behaviour:
The path of the config files should be stored in the framework config file and the function returning the path of config file must not have path hard coded.

This PR fixes the error in default path extraction. When default path is calculated it expects default file path relative to @@@RootDir@@@/configuration.

### DIFF
--- a/configuration/framework_config.cfg
+++ b/configuration/framework_config.cfg
@@ -6,6 +6,11 @@
 
 #------------------------- Settings Directory ----------------------->
 SETTINGS_DIR: ~/.owtf
+CONFIG_DIR: @@@FRAMEWORK_DIR@@@/configuration/
+# These config directory are relative to CONFIG_DIR
+WEB_PLUGIN_CONFIG_DIR: profiles/plugin_web/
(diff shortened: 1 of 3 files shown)


delta24:
@DarKnight24 I specifically wanted to keep the profiles directory separate.

DePierre:
@DarKnight24 Please update the Travis configuration file accordingly

DarKnight24:
@viyatb I will update this pr by shifting hard coded paths to config file.

viyatb:
@DarKnight24 can you elaborate more on the bug, why and what in the description? It would be useful for future reference.
Btw the changes LGTM.

DarKnight24:
@viyatb I have updated the description.

Issue #836, opened 2017-03-05
KeyError while installing - KeyError: 'WORKON_HOME'


sachinkamath:
The installation of OWTF from ./owtf/owtf/install fails with a KeyError

## Current Behavior

Throws a `KeyError` : 
<pre>
Traceback (most recent call last):
  File "install.py", line 307, in <module>
    installer_status_code = install(sys.argv[1:])
  File "install.py", line 211, in install
    setup_virtualenv()
  File "install.py", line 133, in setup_virtualenv
    if os.path.join(os.environ["WORKON_HOME"], "owtf") == os.environ["VIRTUAL_ENV"]:
  File "/usr/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
KeyError: 'WORKON_HOME'

</pre>
## Possible Solution

Check if the environment variable is set.  The problem with the environment variable is not very visible on googling.
(Is this out of context from OWTF point of view?)

## Steps to Reproduce (for bugs)

1. Navigate to installation folder
2. Run the installer

## Your Environment

* Installation method used: Manual
* Version/branch used: develo[
* Operating System and version (like Kali, Debian, ArchLinux, etc): ArchLinux
* pip/setuptools version: pip 9.0.1
* sudo access: Yep :)

viyatb:
@sachinkamath you are not using the latest develop. This `    installer_status_code = install(sys.argv[1:])` was removed sometime back.

sachinkamath:
@viyatb Pretty sure I am on the latest develop :)

![screenshot from 2017-03-06 19-38-34](https://cloud.githubusercontent.com/assets/9546091/23613267/c25af198-02a4-11e7-9bb6-bc4c66d9a14c.png)

viyatb:
@sachinkamath you must've updated after this. The lines on your stacktrace do not exist in the latest develop.

sachinkamath:
@viyatb Yes. After I saw your message, I did update and had the same error. That's why I commented.

viyatb:
@sachinkamath can you post the stacktrace from the updated installation?

sachinkamath:
@viyatb, I pulled #837 from @DarKnight24 's repo and it seemed to have worked. Ran the install again after cleaning up my `.zshrc` and  `git reset --hard HEAD~1` and here is the stacktrace : 

<pre>
Traceback (most recent call last):
  File "install.py", line 304, in <module>
    install(sys.argv[1:])
[...]

PR #838, opened 2017-03-06
Fixed virtualenv conflict with python3

sachinkamath:
## Description

Fixes issue #836
## Related Issue

#836 

## Motivation and Context

This PR fixes an issue with virtualenv when the default python env is 3+
## Reviewers

@DarKnight24 @viyatb 
## How Has This Been Tested?

* Tested on Arch Linux.

## Types of changes

- [ ] Other

### Checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/owtf/owtf/838)

### DIFF
--- a/install/install.py
+++ b/install/install.py
@@ -113,10 +113,13 @@ def finish():
 
 
 def setup_virtualenv():
-    Colorizer.info("[*] Seting up virtual environment named owtf...")
+    Colorizer.info("[*] Setting up virtual environment named owtf...")
+    # If /usr/local/bin/virtualenvwrapper.sh doesn't exist, create a symlink from /usr/bin/
+    if not os.path.isfile('/usr/local/bin/virtualenvwrapper.sh'):
+        run_command('sudo ln -s /usr/bin/virtualenvwrapper.sh /usr/local/bin/virtualenvwrapper.sh >/dev/null 2>&1;')
     # sources files and commands
     source = 'source /usr/local/bin/virtualenvwrapper.sh'
-    setup_env = 'cd $WORKON_HOME; virtualenv -q --always-copy -p %s owtf >/dev/null 2>&1;'\
+    setup_env = 'cd $WORKON_HOME; virtualenv2 -q --always-copy -p %s owtf >/dev/null 2>&1;'\
         ' source owtf/bin/activate' % sys.executable
     dump = '%s -c "import os, json;print json.dumps(dict(os.environ))"' % sys.executable
 
--- a/install/install.py
+++ b/install/install.py
@@ -114,13 +114,21 @@ def finish():
 
 def setup_virtualenv():
     Colorizer.info("[*] Setting up virtual environment named owtf...")
+
+    # Workaround for systems where the default Python Interpreter is 3.0+
+    venv_var = 'virtualenv2'
+    if not os.path.isfile('/usr/bin/virtualenv2'):
+        venv_var = 'virtualenv'
+
     # If /usr/local/bin/virtualenvwrapper.sh doesn't exist, create a symlink from /usr/bin/
     if not os.path.isfile('/usr/local/bin/virtualenvwrapper.sh'):
         run_command('sudo ln -s /usr/bin/virtualenvwrapper.sh /usr/local/bin/virtualenvwrapper.sh >/dev/null 2>&1;')
     # sources files and commands
     source = 'source /usr/local/bin/virtualenvwrapper.sh'
-    setup_env = 'cd $WORKON_HOME; virtualenv2 -q --always-copy -p %s owtf >/dev/null 2>&1;'\
-        ' source owtf/bin/activate' % sys.executable
+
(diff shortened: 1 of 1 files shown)


install/install.py
```diff
@@ -113,10 +113,12 @@ def finish():
 
 
 def setup_virtualenv():
-    Colorizer.info("[*] Seting up virtual environment named owtf...")
+    Colorizer.info("[*] Setting up virtual environment named owtf...")
+    # If /usr/local/bin/virtualenvwrapper.sh doesn't exist, create a symlink from /usr/bin/
+    run_command('sudo ln -s /usr/bin/virtualenvwrapper.sh /usr/local/bin/virtualenvwrapper.sh >/dev/null 2>&1;')
```
viyatb: You don't need to do this every time. The soft link creation should only be done when the `/usr/local/bin` path is not found.

sachinkamath: @viyatb , ln -s will only create the symlink if it doesn't exist. If the file/symlink doesn't exist, it will quietly move on without making any change. Also, installation is only a one-time process and I don't think checking if the file exists or not is necessary.

sachinkamath:
@viyatb Added the check as you suggested.

DePierre:
@sachinkamath You broke Travis. Please make sure that you don't break the CI

sachinkamath:
@DePierre @viyatb Looks like `virtualenv2` is an Arch specific thing and does not seem to exist in distros where py2 ships as the default interpreter. I added an extra check to handle such cases and Travis seems to be passing.

install/install.py
```diff
@@ -113,11 +113,21 @@ def finish():
...
+    # If /usr/local/bin/virtualenvwrapper.sh doesn't exist, create a symlink from /usr/bin/
+    if not os.path.isfile('/usr/local/bin/virtualenvwrapper.sh'):
+        run_command('sudo ln -s /usr/bin/virtualenvwrapper.sh /usr/local/bin/virtualenvwrapper.sh >/dev/null 2>&1;')
     # sources files and commands
     source = 'source /usr/local/bin/virtualenvwrapper.sh'
-    setup_env = 'cd $WORKON_HOME; virtualenv -q --always-copy -p %s owtf >/dev/null 2>&1;'\
-        ' source owtf/bin/activate' % sys.executable
+
+    # Workaround for systems where the default Python Interpreter is 3.0+
+    if not os.path.isfile('/usr/bin/virtualenv2'):
```
DePierre: @viyatb Do you think that this check is enough?

install/install.py
```diff
@@ -113,11 +113,21 @@ def finish():
...
+        setup_env = 'cd $WORKON_HOME; virtualenv -q --always-copy -p %s owtf >/dev/null 2>&1;'\
+            ' source owtf/bin/activate' % sys.executable
+    else:
+        setup_env = 'cd $WORKON_HOME; virtualenv2 -q --always-copy -p %s owtf >/dev/null 2>&1;'\
```
DePierre: It seems very redundant to copy/paste the line twice when the only difference is the additional `2` after `virtualenv`.

Could you instead use `%s` or `.format()`?

install/install.py
sachinkamath: @DePierre I have access to Ubuntu, a Fedora and an Arch box. Looks like virtualenv2 is only applicable on Arch. In all the other distros, I could find `virtualenv` in `/usr/bin`.

sachinkamath:
@DePierre Updated the PR.

DarKnight24:
@sachinkamath I don't understand the use of virtualenv2. The command should create the virtual environment irrespective of python it is using. The problem you had faced due to the wrong path of `virtualenvwrapper.sh` file. This is different in different distros.

viyatb:
@DarKnight24 you are right. @sachinkamath there is no need to do all these checks. We can directly call `virtualenv --python=python2.7` to do this. Look at the virtualenv docs for this.

viyatb:
*[install/install.py, line 124 at r2](https://reviewable.io:443/reviews/owtf/owtf/838#-KebNYpcAWAEA5KKnhaY-r2-124:-KeicRFo0TNLIsnc3St6:bbjxxlv) ([raw file](https://github.com/owtf/owtf/blob/a2ad8ba0433efddcf1a60f9e6572870d1de6227a/install/install.py#L124)):*
<details><summary><i>Previously, sachinkamath (Sachin S. Kamath) wrote…</i></summary><blockquote>

@DePierre I have access to Ubuntu, Fedora and an Arch box. Looks like virtualenv2 is only applicable on Arch. In all the other distros, I could find `virtualenv` in `/usr/bin`.
</blockquote></details>

@sachinkamath You can directly call `virtualenv --python=python2.7` to do this. Look at the virtualenv docs for this.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/owtf/owtf/838)*

viyatb:
Closed in https://github.com/owtf/owtf/commit/81bdca68390a8fa8540e074d6e2296e29ea88f2a.

sachinkamath:
@viyatb Sweet, did not know that. Thanks :)

PR #840, opened 2017-03-10
[virtualenv] Force check for Python virtualenv before any OWTF imports

viyatb:
This PR forces OWTF to check if the Python virtualenv is activated before importing any OWTF modules in the main script.

cc: @DePierre @7a @DarKnight24 feedback please! :)

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -96,20 +96,6 @@ def print_version(root_dir, commit_hash=False, version=False):
         pass
 
 
-def check_if_virtualenv_python():
-    """Checks if the script is running inside a virtualenv or not
-    Stolen from http://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv
-    - Inside a virtualenv, sys.prefix points to the virtualenv directory,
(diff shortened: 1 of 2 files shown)


DarKnight24:
@viyatb Have a look at [this](http://jelly.codes/articles/python-virtualenv-from-within-python/).

DePierre:
@viyatb The code looks fine to me :) (I haven't tested it though)

Quick suggestion: Could we move all code related to dependency check/version check/platform check to `__init__.py`? That would reduce the noise in `owtf.py` and other owtf files IMO.

viyatb:
@DePierre yes, good idea.
@DarKnight24  did you mean that we can use the virtualenv API in `install/install.py`?

DarKnight24:
@viyatb Yes

viyatb:
@DarKnight24 does it have any limitations?

DarKnight24:
@viyatb No, we simply need to run `execfile(os.path.join(venv_dir, "bin", "activate_this.py"))` before importing any dependencies which are installed in virtualenv.

Issue #725, opened 2016-08-23
follow fhs structure, aka make it distro friendly


blshkv:
Hi, I'm trying to port the tool to pentoo and have some mess to clean up.

You keep system (database) configuration files in user's ~/.owtf folder and write results to a current folder which becomes system (read-only, /usr/lib64/owtf/owtf_review) after the tool is installed. It makes it difficult to preinstall and preconfigure it.

The tool should be able to start without ~/.owtf folder, read all configuration from system location and output all results to ~/.owtf.

Ideally, you should make it pip compatible and get rid of all *install shell hacks.

FYI: http://www.pathname.com/fhs/pub/fhs-2.3.html

7a:
Hi Anton,

Thanks for porting OWTF to Pentoo!
The idea of the OWTF output files is to generate them on a test basis, for
example :

cd 2016.08.Some_Client
/path/to/owtf.py

The files are saved on the current directory, which is :

2016.08.Some_Client/owtf_review

Then, you got another pentest, so you run OWTF from that other folder. This
way the OWTF output files are nicely saved on the relevant pentest
directory.

Users should never run owtf from the OWTF directory, maybe we should add a
warning for this  thoughts?

Makes sense?

Thank you and keep us posted!

Abe

blshkv:
ok, I may be wrong about the output, but I found the following:
```
/framework/config/framework_config.cfg:
OUTPUT_PATH: owtf_review
```
So it is a current directory probably. Well, it would be good to have a warning that it will be a directory created, but I still find it is a bad behaviour. During a test, I could be in any directory actually. 

I've managed to convinced authords of few tools such as sqlmap (https://github.com/sqlmapproject/sqlmap/issues/746) and arachni (https://github.com/Arachni/arachni/issues/474)

So we keep configs in /etc/tool/* and the output in ~/.tool/* by default.

delta24:
OWTF installation is primarily there to support Kali 2.x and the current approach works well. If you really need to tweak settings before installing to Pentoo, you can create a simple script which modifies the default **OUTPUT_PATH** to something like `~/.owtf/review/`.

blshkv:
That's exactly my point. The tool should not be attached to a specific distro. You faced some problems already already after Kali moved all tools from /pentest location for example.

Also, the installation scripts are ugly. You should never install any packages from your tool and should never run "sudo". Leave it to a packet manager of each distro.

[...]

PR #843, opened 2017-03-16
Activate virtualenv if not before running OWTF

DarKnight24:
Now OWTF can be executed without activating the virtualenv.

## Description

This PR checks whether virtualenv is activated or not, if not then activates it before running OWTF.
 
## Reviewers

@viyatb @DePierre 

## How Has This Been Tested?

By running owtf in different scenarios.
 
## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/framework/utils.py
+++ b/framework/utils.py
@@ -96,20 +96,6 @@ def print_version(root_dir, commit_hash=False, version=False):
         pass
 
 
-def check_if_virtualenv_python():
-    """Checks if the script is running inside a virtualenv or not
-    Stolen from http://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv
-    - Inside a virtualenv, sys.prefix points to the virtualenv directory,
-    and sys.real_prefix points to the "real" prefix of the system Python (often /usr or /usr/local or some such).
-
-    Outside the virtualenv, sys.real_prefix does not exist.
-    """
-    if not hasattr(sys, 'real_prefix'):
-        print("[-] It seems that virtualenv has not been activated!\n")
-        print("[-] Please run: \t source ~/.%src; workon owtf\n" % os.environ["SHELL"].split(os.sep)[-1])
-        sys.exit(1)
-
-
 class FileOperations(object):
 
     @staticmethod
--- a/owtf.py
+++ b/owtf.py
@@ -6,19 +6,51 @@ In charge of processing arguments and call the framework.
 
 from __future__ import print_function
 
+import json
+import logging
 import os
 import sys
-import logging
+import subprocess
+import traceback
+
+"""Checks if the script is running inside a virtualenv or not
+Stolen from http://stackoverflow.com/questions/1871549/python-determine-if-running-inside-virtualenv
+- Inside a virtualenv, sys.prefix points to the virtualenv directory,
+and sys.real_prefix points to the "real" prefix of the system Python (often /usr or /usr/local or some such).
+
+Outside the virtualenv, sys.real_prefix does not exist.
+"""
+if not hasattr(sys, 'real_prefix'):
+    print("[-] Activating virtualenv 'owtf' ...")
+    # This check ensures that virtualenvwrapper.sh file has been activated
+    if os.environ.get("WORKON_HOME") is None:
+        shell_rc_path = os.path.join(os.environ["HOME"], ".%src" % os.environ["SHELL"].split(os.sep)[-1])
+        dump = '%s -c "import os, json;print json.dumps(dict(os.environ))"' % sys.executable
+        pipe = subprocess.Popen(['/bin/bash', '-c', 'source %s >/dev/null 2>&1; %s' % (shell_rc_path, dump)],
+                                stdout=subprocess.PIPE)
+        env = json.loads(pipe.stdout.read())
+        # Update the os environment variable
+        os.environ.update(env)
+
+    try:
+        activate_script = os.path.join(os.environ["WORKON_HOME"], "owtf", "bin", "activate_this.py")
+        if os.path.exists(activate_script):
+            execfile(activate_script, dict(__file__=activate_script))
+        else:
+            print("[-] Unable to find virtualenv 'owtf'. Please rerun the install script")
+            sys.exit(1)
+    except:
+        traceback.print_exc()
+        print("Unable to activate virtualenv. Please report the error on https://github.com/owtf/owtf")
+        sys.exit(1)
 
 from framework.dependency_check import verify_dependencies
 from framework.core import Core
 from framework.dependency_management.component_initialiser import ComponentInitialiser, DatabaseNotRunningException
 from framework.dependency_management.dependency_resolver import ServiceLocator
 from framework import update
-from framework.utils import check_if_virtualenv_python
 from framework.lib.cli_options import usage, parse_options, parse_update_options
 
-check_if_virtualenv_python()
 verify_dependencies(os.path.dirname(os.path.abspath(sys.argv[0])) or '.')
 
 


DePierre:
*My comment below might be a rant about OWTF's installation process. If you believe that my arguments are wrong, feel free to ignore me. Right now, I believe that they are justified.*

I see more and more code related to OWTF's installation, which contain hackish calls to system, exec and whatnots. It doesn't make me feel comfortable personally.

I have no experience in installation process but I believe that we reached a point (or reached months, years ago) where we are stepping out of what OWTF should be doing. For instance:

+ The user should decide whether he wants to run OWTF in a virtual environment or not.
+ There is a problem with the user SSL configuration? OWTF should not try to fix it.
+ Kali broke something? OWTF shoud not try to fix it.
+ pip broke something? OWTF should not try to fix it.

OWTF is not supposed to try to fix the user's system or other packages' issues. We already have a lot to fix in our own codebase already IMO. OWTF should provides a clean installation process where it only installs the python PyPi packages it needs, creates the directories it needs and nothing else. Skipfish is not present? Not OWTF's problem to fix. W3AF does not work? Not OWTF's problem to fix. PostgreSQL is not started? Not OWTF's problem.

Is there any one who has experience with installation process and could provide meaningful input? I remember that https://github.com/owtf/owtf/issues/725 (opened by @blshkv) seemed to contain good arguments and recommendations on how the installation process should go. I would like to have an outside point of view about OWTF's current installation process.

I would like to strip all hackish code like the one provided in this PR (nothing personal @DarKnight24) from the OWTF code base. I would agree on creating a new repository containing all hackish code for fixing Kali, starting postgresql, etc. that the user may want to run if he wants to *fix* his Kali distribution.

I know it's old but I remember the time where I could simply `git clone owtf.git && pip install -r install/owtf.pip && python install/install.py` on my Archlinux. And where I did not need to be root!

viyatb:
@DePierre in short, I agree with you. But there are a few considerations:

* re: virtualenv - Kali changed the permissions on the global Python packages directory. You can no longer update/upgrade packages already installed (because they would break the existing bundled tools).
This is the sole reason we need virtual environment for OWTF.
* re: SSL problems - I agree. They were a short term fix and meant to be removed (the release cycle at that time was not rolling, hence they were kept.)
* re: Kali broke - the tools adapt to the environment, otherwise it would be unusable for most users. Some feature might break the functionality for a small subset of tools and that may be a low priority for Kali maintainers to fix. What then?
* re: Pip broke something - again the same thing as above. Usability is the priority here for us.
* re: system scripts: This is why we have separate scripts which help fix these issues in the `scripts/` folder and not the `framework`. However I strongly disagree on decoupling those scripts from the main repository.

Let me know what you think. @7a can you pitch in? :)

DePierre:
@viyatb Thanks for providing some feedback. Just to be more realistic, the current situation about having more and more code related to install/env/dependency inside OWTF's source code is still manageable. However, it seems to get worse so I would rather be alarmist now than hopeless later.

I am surely overreacting.

> - re: virtualenv - Kali changed the permissions on the global Python packages directory. You can no longer update/upgrade packages already installed (because they would break the existing bundled tools).
> This is the sole reason we need virtual environment for OWTF.

I was not aware of this and it makes some sense from Kali's point of view. Do we know how other tools proceed? Are they all moving to virtualenv?

> - re: Kali broke - the tools adapt to the environment, otherwise it would be unusable for most users. Some feature might break the functionality for a small subset of tools and that may be a low priority for Kali maintainers to fix. What then?

If it is Kali's fault, then we do nothing but let the users complain to Kali. If Kali disagrees, then too bad for the users but we can still provide them with links to work-arounds and scripts, **independent** from OWTF normal installation. Which, to be honest, is what we have started to do for a while now (like asking the user if he wanted OWTF to fix the SSL configuration for postgresql, instead of silently patching it for him)

> - re: Pip broke something - again the same thing as above. Usability is the priority here for OWTF.

My position is that OWTF **should not** have to fix other project's bugs. If pip broke something, then it's pip's responsibility to revert their changes or provide the new way to do what we were doing in the past.

We already have the heavy burden of maintaining our list of plugins working (and not all of them are working). And it's a lot to do! My example with w3af might have been misleading in the sense that OWTF **should** keep its w3af configuration up-to-date to work with W3AF's latest version. However, OWTF **should not** try to hot-patch w3af's code to fix a w3af bug.

> - re: system scripts: This is why we have separate scripts which help fix these issues in the scripts/ folder and not the framework. However I strongly disagree on decoupling those scripts from the main repository.

You are right, it's already decoupled if they are in a dedicated directory and I don't mind keeping them there. I might have gone too far with the dedicated repository idea. What I want is to get rid of all the python code inside OWTF's source code that are related to dependency checks, virtualenv checks, etc.

**To rephrase**, the current situation is not as bad as I pictured it. There is indeed a dedicated directory where temporary fixes have been written. However, the trend I am seeing is more and more code inside OWTF that are related to checking installation things (like whether dependencies have been installed or virtualenv has been enabled) that should not be part of the core code.

How could the following code (from the PR) belong to OWTF's code base?

```python
shell_rc_path = os.path.join(os.environ["HOME"], ".%src" % os.environ["SHELL"].split(os.sep)[-1])
dump = '%s -c "import os, json;print json.dumps(dict(os.environ))"' % sys.executable
pipe = subprocess.Popen(['/bin/bash', '-c', 'source %s >/dev/null 2>&1; %s' % (shell_rc_path, dump)], stdout=subprocess.PIPE)
env = json.loads(pipe.stdout.read())
# Update the os environment variable
os.environ.update(env)
```

(Again, nothing personal @DarKnight24. I am pointing out that OWTF embeds more and more snippet like yours that do not belong to OWTF's code base. OWTF is a python framework to provide an easy-way for users to follow pen-testing guides. OWTF should not be a freak of fake python code that is in fact bash code just to do something that we shouldn't)

blshkv:
Guys, 
By looking at your comments I got an impression that you have finally realised that you are getting too far. So far, that eventually your script will be downloading a specific ISO image of kali image first, patching all stuff and acting like your tool is the only tool a user would run (and breaking other tools potentially). This is obvious a wrong path.

You should care less whether user's env it is docker, virtualenv, kali or any other linux distro. 
Please provide a *standard* setup.py (https://docs.python.org/3/install/) and let distro maintainers do their job.

From the other side, owtf *should*:
 - expect that it will be installed in a system directory
 - able to run under regular user's privilege (unless there is a really good reason)
 - pre-install required configuration files in /etc file
 - create ~.owtf/* folders/files during runtime if they are missing (instead of crashing).
 - supervise local configuration files in the local ~./owtf folder
 - redirect all output to that local directory *by default*
 - handle other missing dependencies properly (just display a warning and disable the functionality)
The list can go on.

This is how it works for thousand tools up there.

DePierre:
@blshkv One specificity about OWTF that could justify why we took the wrong decision of focusing on Kali and getting too far is that OWTF relies on 10, 20 even 30 different security tools (tools, not libraries or packages. actual tools that should be used by human users in the first place) and Kali is among the few distributions that provides them pre-installed (others being BlackArch).

It's therefore understandable to couple OWTF and Kali. I agree with you, it's the wrong path, but I thought I could give some insight about why.

viyatb:
@blshkv you are missing the main point: OWTF is the glue for the thousands of tools you mentioned.  Its a framework first and then a tool.

@DePierre I can start removing the cruft from the codebase. I was thinking of creating a branch from develop to rewrite, modify, refactor OWTF inside out.

blshkv:
@viyatb I'm pretty sure I'm not missing anything ;-) owtf is just one of thousands frameworks in my system, however it is trying to patch other frameworks (such as w3af) and modify system files without my permission. I have to rip off these malicious functions from each version of the tool and this is a lot of work.

viyatb:
Closing the PR according to the above discussion.

PR #861, opened 2017-07-30
Fix issue when target url contains reserved characters

csk:
Fix issue when target url contains reserved characters (like questionmark)

## Description

There is an issue with the fileserver and the encoding when a target contains special characters. E.g: http://192.168.0.14/mutillidae/index.php?page=user-info.php

This tiny commit fixes it :)

### DIFF
--- a/framework/interface/reporter.py
+++ b/framework/interface/reporter.py
@@ -7,6 +7,7 @@ The reporter module is in charge of producing the HTML Report as well as
 import cgi
 
 from tornado.template import Loader
+from tornado.escape import url_escape
 
 from framework.dependency_management.dependency_resolver import BaseComponent
 from framework.dependency_management.interfaces import ReporterInterface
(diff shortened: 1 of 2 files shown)


viyatb:
@csk the PR looks good. I'll merge after testing this later today.

viyatb:
@csk did you test it? I cannot replicate the fix.

csk:
Yes i did! The problem is that if the target contains reserved characters, the plugin output files are stored in a directory named as by the url.

screen shot 2017-08-01 at 1 36 02 am

The link generated by the reported, if nos correctly escaped, is rendered like:

screen shot 2017-08-01 at 1 33 14 am

The link in the screenshot above is handled by the browser as a url with query params if not escaped correctly and the webserver receives the request and redirects wrongly to an unexistant resource:

screen shot 2017-08-01 at 1 40 21 am

viyatb:
@csk I meant the fix is incorrect. You need to modify `includes/src/Report/Table.jsx` here https://github.com/owtf/owtf/blob/develop/includes/src/Report/Table.jsx#L59.
The fix is `encodeURIComponent(obj['output_path'])`

csk:
I got it! Totally agree with you, the path should not be encoded until displaying it on the web as link. I will fix it and PR.

csk:
@viyatb i just found out that there is non-React stuff which is rendered server side which is also affected.

https://github.com/owtf/owtf/blob/master/framework/interface/templates/poutput/command_dump.html#L8 and https://github.com/owtf/owtf/blob/master/framework/interface/templates/poutput/command_dump.html#L19

Maybe we could pass a uriEncodedFilePath to the template. Do you agree?

viyatb:
@csk how is that affected on the UI? I don't see the point of encoding this in the template. These are being deprecated.

PR #868, opened 2017-08-03
Update reporter.py

Tikam02:
sys.exit(-1) 
the main issue was if tha data = json.load(data_file) if the file was empty and read from index 0 or 1 it shows out of index.

reporter py
screenshot from 2017-08-03 05-59-51


DePierre:
@Tikam02 Thanks for the PR. However, it is completely wrong. What your changes does is to force `reporter.py` to exist in **all** circumstances. It will never go past line 17 and will never create the HTML report.

Please take your time and read through the items we discussed yesterday. This is not a proper fix.

Issue #695, opened 2016-05-27
[webui] Redesign the web interface


viyatb:
This is just the placeholder issue for web interface redesign.

> X-reboot issue: https://github.com/owtf/reboot/issues/18
- Reduce clicks and have a more natural flow ⇒ This refers to the fact that the current interface is not user-friendly, and can be modified to increase productivity while reducing redundancies and improving consistency.
- The main OWTF report uses elements known as accordions to represent findings of the plugins. The accordion element is not scalable: it is not a user-friendly technique to represent hundreds of rows of plugins.

The goal here is to come up with an interface that is user-friendly, scalable, and maintainable.
- CSS standards: Currently OWTF uses a heavy-CSS framework, Twitter Bootstrap to render elements. Twitter Bootstrap includes many unnecessary elements and is so heavy-handed, so prescriptive, that it is actually a nuisance to modify. It’s easier to use the defaults.
- Also write everything in SASS (Syntactically Awesome Stylesheets), reducing redundancy and development time.and increasing consistency. SASS is very easy to write and is a more maintainable alternative.
- Use compass (SASS -> CSS compiler) to compile everything (concatenate) the stylesheets and minify them whenever regenerating.

delta24:
> Shifted here because it is a subset of the main issue.

At the moment you need 3 clicks + 1 nasty horizontal mouse move from left to right to rank ANY plugin.
We need to:
- Put the plugins somehow into the plugin group banners (-1 click)
- you know the plugins that have been run under that test, we could abbreviate with [ E P S A ] ← External, Passive, Semi-Passive, Active :)
- Put the buttons to click on the plugins [ E P S A ] directly ABOVE the ranking buttons (decreased horizontal mouse trip)
- [ ] Optimise UI of the report
- [ ] ZAP/Burp like transaction log viewer
- [ ] Replace accordions to something more easier to navigate

viyatb:
@saganshul there are a couple of bugs in the new interface:

- [ ] The filters do not work.
- [ ] The proxy transactions should ideally be colored when "selected" => highlight the row.
- [ ] The transactions view does not have pagination.
- [ ] Browse files do not give the correct path from the file server (assigned to me).

saganshul:
@viyatb In first point, which filters are you talking about?

viyatb:
@saganshul the advanced filters in the report.

saganshul:
@viyatb I checked advance filters are working correctly. Can you please specifically tell me which filter is not working?

[...]

PR #870, opened 2017-08-04
Adds export report funtionality

saganshul:
Uses [docxtemplater](https://github.com/open-xml-templating/docxtemplater) to create export report functionality.

## Description
Rest API - /api/targets/<target_id>/export/
Google DOC (Input Template)- https://docs.google.com/document/d/1lVWs0hV9SROAaqbReV0FJHfKphrv1dH7nS9K11Vi3X4/edit?usp=sharing

## Related Issue
#695 

## Reviewers
@viyatb @7a 

## How Has This Been Tested?
Tested on Google Chrome browser.

screenshot from 2017-08-04 19-52-54
screenshot from 2017-08-04 19-53-05

## Types of changes

- [ ] Other


framework/interface/api_handlers.py
```diff
@@ -842,3 +843,65 @@ def delete(self, error_id=None):
...
+                mappings = self.get_component("mapping_db").GetMappings(self.get_argument("mapping", None))
+            else:
+                mappings = None
+
+            # Get test groups as well, for names and info links
+            test_groups = {}
+            for test_group in self.get_component("db_plugin").GetAllTestGroups():
+                test_group["mapped_code"] = test_group["code"]
+                test_group["mapped_descrip"] = test_group["descrip"]
+                if mappings:
```
DePierre: Could be shortened:
```python
if mappings and test_group['code'] in mappings:
    test_group["mapped_code"] = mappings[test_group['code']][0]
    test_group["mapped_descrip"] = mappings[test_group['code']][1]
```
Also, it would be more readable to name `[0]` and `[1]`, for instance:
```python
code, description = mappings[test_group['code']]
test_group["mapped_code"] = code
test_group["mapped_descrip"] = description
```

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,65 @@ def delete(self, error_id=None):
...
+                    try:
+                        test_group["mapped_code"] = mappings[test_group['code']][0]
+                        test_group["mapped_descrip"] = mappings[test_group['code']][1]
+                    except KeyError:
+                        pass
+                test_groups[test_group['code']] = test_group
+
+            vulnerabilities = []
+            for key, value in grouped_plugin_outputs.iteritems():
```
DePierre: Just a note: `iteritems` does not exist in python 3.x. If OWTF is planing on supporting python3.x, it would be time to stop adding more incompatible code. Would `items` be enough? Or will it add to much overhead in python2.x?

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,65 @@ def delete(self, error_id=None):
...
+                obj = test_groups[key]
```
DePierre: Is there a need for the temporary variable (for creating a copy and not a reference)? Or would the following work as well?
```python
for key, value in grouped_plugin_outputs.items():
    test_groups[key]['data'] = value
    vulnerabilities.append(test_groups[key])
```

Also in your current version, are we sure that `test_groups[key]` exists?

framework/utils.py
```diff
@@ -32,6 +32,32 @@ def is_internal_ip(ip):
     parsed_ip = IPAddress(ip)
     return parsed_ip.is_private
 
+def get_rank(user_rank, owtf_rank):
```
DePierre: Are you sure that such function is not already available? Either in OWTF or even in PTP?

DePierre: Apparently no, there is no other reference where the ranking scale is defined (except hardcoded in https://github.com/owtf/owtf/blob/master/framework/interface/templates/plugin_report.html)

DePierre:
@saganshul Thank you very much for implementing that very useful feature!

One thing that bothers me a little bit is to understand how easy/painful it would be to support other templates. In your opinion, do your changes allow supporting multiple report templates? If not, wouldn't it be better to spend a little more time on the feature and include that thought in the code?

I would personally imagine a configuration file defining the required information in a report (vuln title, description, ranking, etc.). Then I would imagine OWTF supporting at least 2 or 3 templates (one with highest rating first for instance, one ordered by plugin groups, etc.). And finally I would imagine having the ability as a user to select among the different templates.

viyatb:
@DePierre Anshul and I thought about creating a repository of community supported templates so that users can have the choice as many as variations they want. The configuration file would not be necessary since the user can trivially move around the parts in the DOCX/ODT file.

DePierre:
@viyatb How will OWTF fill the fields in the report templates? Is there a well-defined list of elements that are required, optional, and so on for OWTF templates?

I am 100% for that community repository writing the multiple templates. It's really an awesome idea! But then it would require very strong documentations about what can/should/may/might be included in the template.

Going a bit further, it might be required to implement the report generation using a plugin system, since it will have to deal with potentially 10s of different templates. As far as I can read in this PR, everything is hardcoded for one template. This is a great PoC but not something that is maintainable and can be improved easily for multiple formats.

I am not trying to diminish any work already done here. However more work should be put in the design of the feature to be both easily maintainable and improved. It should also be properly documented, which doesn't seem to be the case in this PR (regarding what fields are available, required, optional, etc.).

viyatb:
@DePierre yeah, documentation :sighs: lol. All your suggestions are on point and is going to done in iterative commits. 👍

viyatb:
@saganshul the JS part LGTM.

framework/utils.py
saganshul: No there is not such a function exists.

saganshul:
@viyatb @DePierre PR updated. Please review changes :)

saganshul:
I changed commit message because currently it is basic implementation.
TODO:
- Add multi template support(Using community supported templates)(that would be very easy)
- Support other formats like PDF, HTML etc.

@DePierre The same API will work for all templates.  API throws all the information it has in OWTF DB about scan. So we just need to write template. Same API works for all template. Also multiple formats can be supported since what we can convert docx to different formats. 
I am not getting what do you mean by this `everything is hardcoded for one template`. There is nothing hard coded all data is coming from API (Template only needs to know keys and JSON structure). Please correct me if I am wrong somewhere or if you are expecting something different.

DePierre:
@saganshul Sorry maybe I was not clear in my previous comment.

Regarding the multiple templates, I meant that currently, only one .docx template is supported (the one you linked) and that currently, the PR does not easily allow using another one. For instance, there could be a second template that sorts the findings by plugin type (web active first, web semi-passive second, web passive third, etc.). In that case, it will require some changes in the feature to easily support that second template. Maybe I am missing something that I am not seeing.

Regarding the *everything is hardcoded for one template*, you just mentioned the issue: **Template only needs to know keys and JSON structure**. In other, if I want to add the duration time for each plugin, I will have to modify the feature to add one additional key. Or is it exposed? Once again, I am maybe missing something that I am not seeing.

This last remark is also related to the lack of documentation. What does OWTF offer in terms of data? You say **API throws all the information it has in OWTF DB about scan**. In the PR, I see that it offers the vulnerability title, its data, etc. But I only understand that by reading the source code or know the inner working of its DB. For instance, with your PR, is it possible to know if a ranking is coming from OWTF or the user? I would believe that we would need to clearly define all the data that one would need in a report and expose all of them. Then, each template would have to follow that documentation and pick the data it needs.

Does it make more sense?

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,71 @@ def delete(self, error_id=None):
...
+        """
+        REST API - /api/targets/<target_id>/export/ returns JSON(data) for template.
+        """
+        if not target_id:
+            raise tornado.web.HTTPError(400)
+        try:
+            filter_data = dict(self.request.arguments)
+            plugin_outputs = self.get_component("plugin_output").GetAll(filter_data, target_id=target_id, inc_output=True)
+            # Group the plugin outputs to make it easier in template
+            grouped_plugin_outputs = {}
```
DePierre: I believe that this could be shortened with:
```python
grouped_plugin_outputs = {poutput['plugin_code']: [] for poutput in plugin_outputs}
```

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,71 @@ def delete(self, error_id=None):
...
+            result = self.get_component("target").GetTargetConfigForID(target_id)
+            result["vulnerabilities"] = vulnerabilities
+            result["time"] = strftime("%Y-%m-%d %H:%M:%S", gmtime())
+
+            if result:
+                self.write(result)
+            else:
+                raise tornado.web.HTTPError(400)
+
+        except exceptions.InvalidTargetReference as e:
```
DePierre: Could we reduce the scope of the try/except statement? It is usually recommended to have the smallest scope for try/except.

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,71 @@ def delete(self, error_id=None):
...
+        if not target_id:
+            raise tornado.web.HTTPError(400)
+        try:
+            filter_data = dict(self.request.arguments)
+            plugin_outputs = self.get_component("plugin_output").GetAll(filter_data, target_id=target_id, inc_output=True)
+            # Group the plugin outputs to make it easier in template
+            grouped_plugin_outputs = {}
+            for poutput in plugin_outputs:
+                if grouped_plugin_outputs.get(poutput['plugin_code']) is None:
+                    # No problem of overwriting
```
DePierre: I am not sure to understand the the point of this comment. What is overridden? Maybe I am not understanding the code snippet but I believe that the line below this comment does not override anything.

framework/interface/api_handlers.py
```diff
@@ -842,3 +843,71 @@ def delete(self, error_id=None):
             self.get_component("db_error").Delete(error_id)
         except exceptions.InvalidErrorReference:
             raise tornado.web.HTTPError(400)
+
+class ReportExportHandler(custom_handlers.APIRequestHandler):
+    """
+    Class handling APIs related to export report funtionality.
+    In OWTF, export report funtionality is implemented using docxtemplater.
+    docxtemplater takes two inputs:
```
DePierre: I don't believe that this belongs to the docstring of the API. As far as I can see, docxtemplater is only being used in a JS file that should be abstracted from this API. It does not and should not be a concern for the API whether `docxtemplater` or .docx format is used or not.

The docstring should focus on what the API does. As far as I can tell, it fetches data from the DB to prepare some JSON structure that contain information about a target, plugins run, rankings, etc. What information does it fetch? What does it do with it? What return value does it return?

I would prefer to have a docstring that explains what it does rather than a docstring that explains what another piece of code somewhere else can do with the data it created.

framework/interface/urls.py
```diff
@@ -34,6 +34,7 @@ def get_handlers():
         tornado.web.url(r'/api/targets/([0-9]+)/transactions/hrt/?([0-9]+)?/?$', api_handlers.TransactionHrtHandler, name='transactions_hrt_api_url'),
         tornado.web.url(r'/api/targets/([0-9]+)/poutput/?' + plugin_group_re + '/?' + plugin_type_re + '/?' + plugin_code_re + '/?$', api_handlers.PluginOutputHandler, name='poutput_api_url'),
         tornado.web.url(r'/api/targets/([0-9]+)/poutput/names/?' + plugin_group_re + '/?' + plugin_type_re + '/?' + plugin_code_re + '/?$', api_handlers.PluginNameOutput, name='plugin_name_api_url'),
+        tornado.web.url(r'/api/targets/([0-9]+)/export/?$', api_handlers.ReportExportHandler, name='report_export_api_url'),
```
DePierre: It might befar-fetched but wouldn't it be more modular to have the API set to `/api/targets/{id}/export/{format}` or `/api/targets/{id}/export/{template}` instead?

It would free the `/api/targets/{id}/export/` API that could later default to a specific template/format, or be used to show a dialog box where the user can specify more options.

What do you think?

framework/utils.py
DePierre: As such, it would be better to separate the `get_rank` function from the ranking scale. By splitting the two, the ranking scale could be used somewhere else in the framework, in case another feature needs to translate the rankings.

Also, `get_rank` should check that the index is defined in the dict and default to a default value if it is not. For instance, something like:
```python
return RANKS.get(rank, 'Unknown')
```
The string `Unknown` should be defined somewhere else, or `Unranked` could be used. You get the idea.

 I would suggest specifying what is the return value in the docstring. Also, I am not sure to understand why the function should make the decision of which ranking to use. I believe that this should be done before calling it. Something like:
```python
# From the API
rank_str(max(user_rank, owtf_rank))
```

Finally, the name of the function could be a bit more specific in what it does with something like `rank_human_readable`, `rank_str`. From what I read, the function does not `get` anything, it simply translates a ranking from numerical to string value.

DePierre: Now I am realizing that we don't need such function. Instead, all we need is the dict translating the numerical ranking to a string.

The `get_rank` function can be removed and instead we could do the following:
```python
# In some util file
from ptp.libptp.constants import INFO, LOW, MEDIUM, HIGH
OWTF_UNRANKED = -1
OWTF_PASSING = 0
OWTF_CRITICAL = 5
RANKS = {
    OWTF_UNRANKED: 'Unranked',
    OWTF_PASSING: 'Passing',
    INFO: 'Informational',
    LOW: 'Low',
    MEDIUM: 'Medium',
    HIGH: 'High',
    OWTF_CRITICAL: 'Critical'
}

# In the API
poutput["rank"] = RANKS.get(max(user_rank, owtf_rank), RANKS[OWTF_UNRANKED])

File: framework/interface/urls.py (hunk 9 above)
Comment (saganshul) [inline framework/interface/urls.py]: @DePierre Actually, we have plans to separate the React UI from Tornado API server. Means Render everything using React rather than Jinja2. We will make backend as simple API server. 

Now, as you suggested API should be something like this -> `/api/targets/{id}/export/{template}`  , but backend server will not have any information about templates, as the export functionality is completely client side and only React or UI will have information about available templates. 
 
Use of `/api/targets/{id}/export/` is to just return data from DB, `docxtemplater` will do the rest. I mean we don't need to have seperate api for template, because the same API will work for all template.

Comment (DePierre) [inline framework/interface/urls.py]: @saganshul Thanks for the additional info. I believe you already thought about it and it's fine by me. It was simply a remark.

Did you have time to consider the other review comments I made?

Comment (saganshul) [inline framework/interface/urls.py]: yeah! Just doing that :)

File: framework/interface/api_handlers.py (hunk 5 above)
Comment (saganshul) [inline framework/interface/api_handlers.py]:
```
for poutput in plugin_outputs:
    if grouped_plugin_outputs.get(poutput['plugin_code']) is None:
        grouped_plugin_outputs[poutput['plugin_code']] = []

    poutput["rank"] = get_rank(poutput["user_rank"], poutput["owtf_rank"])
    grouped_plugin_outputs[poutput['plugin_code']].append(poutput)
```

This total thing can't be replaced with 
```
grouped_plugin_outputs = {poutput['plugin_code']: [] for poutput in plugin_outputs}
```

File: framework/utils.py (hunk 4 above)
Comment (saganshul) [inline framework/utils.py]: @DePierre Can I get all ` INFO, LOW, MEDIUM, HIGH, UNRANKED, PASSING, CRITICAL` variable from ptp.

Comment (DePierre) [inline framework/utils.py]: @saganshul ptp does not have `CRITICAL` by design, since it deemed that only user could have the context to rank an issue as `CRITICAL`.

`PASSING` does not exist, since ptp does not evaluate if a report is passing or not. However, `UNKNOWN` is used when no ranking could be found. It should be noted though that `UNKNOWN` is `0`, while OWTF uses `0` for `PASSING`.

See https://github.com/owtf/ptp/blob/develop/ptp/libptp/constants.py

File: framework/interface/api_handlers.py (hunk 5 above)
Comment (DePierre) [inline framework/interface/api_handlers.py]: @saganshul I understand, I misread the indentation and was only considering the first three lines.

Would the following structure be better in your opinion?
```python
from collections import defaultdict
 
grouped_plugin_outputs = defaultdict(list)
for output in plugin_outputs:
    output['rank'] = RANKS.get(max(output['user_rank'], output['owtf_rank']))
    grouped_plugin_outputs[output['plugin_code']].append(output)
```

saganshul:
@DePierre @viyatb I updated the PR.
* Code improvements.
* Multi template support.

TODO:
* API documentation.

viyatb:
@saganshul do you have any screenshot on how you choose a specific template?

framework/constants.py
```diff
@@ -0,0 +1,15 @@
...
+OWTF_UNRANKED = -1
+OWTF_PASSING = 0
+OWTF_INFO = 1
+OWTF_LOW = 2
+OWTF_MEDIUM = 3
+OWTF_HIGH = 4
+OWTF_CRITICAL = 5
+
+# Maps `int` value of ranks with `string` value.
+RANKS = {OWTF_UNRANKED: 'Unranked', OWTF_PASSING: 'Passing', OWTF_INFO: 'Informational', OWTF_LOW: 'Low', OWTF_MEDIUM: 'Medium', OWTF_HIGH: 'High', OWTF_CRITICAL: 'Critical'}
```
viyatb: Maybe format this in pretty print format.

```
RANKS = {
    OWTF_UNRANKED: 'Unranked',
    OWTF_PASSING: 'Passing',
    OWTF_INFO: 'Informational',
    OWTF_LOW: 'Low',
    OWTF_MEDIUM: 'Medium',
    OWTF_HIGH: 'High',
    OWTF_CRITICAL: 'Critical',
    }
```

saganshul:
@viyatb 
screenshot from 2017-08-09 13-12-31

PR #873, opened 2017-08-16
Fixes the Advance filter, mapping filter bug

saganshul:
Fixes the mapping filtering.

## Description
Fix adds mapping object with `/names` APIs which then used to filter mapping on client side only.

## Related Issue
#695 

## Reviewers
@viyatb @7a @DePierre 

screenshot from 2017-08-16 19-51-24
screenshot from 2017-08-16 19-51-26

## Types of changes

- [ ] Other

### DIFF
--- a/framework/db/mapping_manager.py
+++ b/framework/db/mapping_manager.py
@@ -74,6 +74,15 @@ class MappingDB(BaseComponent, MappingDBInterface):
         """
         return self.mapping_types
 
+    def get_all_mappings(self):
+        """Create a mapping between OWTF plugins code and OWTF plugins description.
+
+        :return: Mapping dictionary {code: [mapped_code, mapped_description], code2: [mapped_code, mapped_description], ...}
+        :rtype: dict
+        """
+        mapping_objs = self.db.session.query(models.Mapping).all()
+        return {mapping['owtf_code']: mapping['mappings'] for mapping in self.DeriveMappingDicts(mapping_objs)}
+
     def GetMappings(self, mapping_type):
         if mapping_type in self.mapping_types:
             mapping_objs = self.db.session.query(models.Mapping).all()
--- a/framework/interface/api_handlers.py
+++ b/framework/interface/api_handlers.py
@@ -52,39 +52,28 @@ class PluginDataHandler(custom_handlers.APIRequestHandler):
 class PluginNameOutput(custom_handlers.UIRequestHandler):
     SUPPORTED_METHODS = ['GET']
 
-    def get(self, target_id=None, plugin_group=None, plugin_type=None, plugin_code=None):
+    def get(self, target_id=None):
+        """Retrieve scan results for a target.
+
+        :return: {code: {data: [], details: {}}, code2: {data: [], details: {}} }
+
+        This API doesn't return `output` section as part of optimization.
+        `data` is array of scan results according to `plugin_types`.
+        `details` contains info about `code`.
+
+        """
         try:
             filter_data = dict(self.request.arguments)
-            if plugin_group and not plugin_type:
-                filter_data.update({"plugin_group": plugin_group})
-            if plugin_type and plugin_group and (not plugin_code):
-                if plugin_type not in self.get_component("db_plugin").GetTypesForGroup(plugin_group):
-                    raise tornado.web.HTTPError(400)
-                filter_data.update({"plugin_type": plugin_type, "plugin_group": plugin_group})
-            if plugin_type and plugin_group and plugin_code:
-                if plugin_type not in self.get_component("db_plugin").GetTypesForGroup(plugin_group):
-                    raise tornado.web.HTTPError(400)
-                filter_data.update({"plugin_type": plugin_type, "plugin_group": plugin_group, "plugin_code": plugin_code})
             results = self.get_component("plugin_output").GetAll(filter_data, target_id=int(target_id), inc_output=False)
 
             # Get mappings
-            if self.get_argument("mapping", None):
-                mappings = self.get_component("mapping_db").GetMappings(self.get_argument("mapping", None))
-            else:
-                mappings = None
+            mappings = self.get_component("mapping_db").get_all_mappings()
 
-            ## Get test groups as well, for names and info links
-            test_groups = {}
-            for test_group in self.get_component("db_plugin").GetAllTestGroups():
-                test_group["mapped_code"] = test_group["code"]
-                test_group["mapped_descrip"] = test_group["descrip"]
-                if mappings:
-                    try:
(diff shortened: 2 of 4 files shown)


framework/db/mapping_manager.py
```diff
@@ -74,6 +74,13 @@ def GetMappingTypes(self):
         """
         return self.mapping_types
 
+    def GetAllMappings(self):
```
viyatb: Snake case here.

saganshul: Are we changing case system because all other function names are in `camel case` ?

viyatb: Function names should always be in snake case according to PEP specifications. I am changing every other names when refactoring.

framework/db/mapping_manager.py
```diff
@@ -74,6 +74,13 @@ def GetMappingTypes(self):
...
+        mapping_objs = self.db.session.query(models.Mapping).all()
+        mappings = {}
+        for mapping_dict in self.DeriveMappingDicts(mapping_objs):
+            mappings[mapping_dict["owtf_code"]] = mapping_dict["mappings"]
+        return mappings
```
DePierre: Could this be shortened like the following?
```python
return {mapping['owtf_code']: mapping['mappings'] for mapping in self.DeriveMappingDicts(mapping_objs)}
```

framework/db/mapping_manager.py
DePierre: In addition, a quick docstring to document the purpose of the function would be welcome! :)

framework/interface/api_handlers.py
```diff
@@ -52,38 +52,20 @@ def get(self, plugin_group=None, plugin_type=None, plugin_code=None):
 class PluginNameOutput(custom_handlers.UIRequestHandler):
     SUPPORTED_METHODS = ['GET']
 
-    def get(self, target_id=None, plugin_group=None, plugin_type=None, plugin_code=None):
+    def get(self, target_id=None):
```
DePierre: A quick docstring to document the purpose of the function would be welcome! :)

framework/interface/api_handlers.py
```diff
@@ -52,38 +52,20 @@ def get(self, plugin_group=None, plugin_type=None, plugin_code=None):
...
-                test_group["mapped_descrip"] = test_group["descrip"]
-                if mappings:
-                    try:
-                        test_group["mapped_code"] = mappings[test_group['code']][0]
-                        test_group["mapped_descrip"] = mappings[test_group['code']][1]
-                    except KeyError:
-                        pass
+                test_group["mappings"] = {}
+                if test_group['code'] in mappings:
+                    test_group["mappings"] = mappings[test_group['code']]
```
DePierre: Could this be shortened like the following?
```python
groups = {}
for group in self.get_component("db_plugin").GetAllTestGroups():
    group['mappings'] = mappings.get(group['code'], {})
    groups[group['code']] = group
```

framework/interface/api_handlers.py
saganshul: Docstring added.

saganshul:
@DePierre @viyatb PR updated. Please review.
Thanx @DePierre for suggesting short code snippets.

framework/interface/api_handlers.py
DePierre: @saganshul Thanks for taking the time to add documentation. However I feel that the documentation you added for this API is incorrect in several ways that I would like to present:

+ The rest API path should not be listed in the docstring since both are independent. We could imagine moving API around, which would make this doc outdated
+ How the returned data is being used should not be listed either since it is also independent from the function itself

Instead, I feel that a docstring explaining what information is generated and returned would be better. You started explaining that on the second line with `This API returns information about a target scan present in OWTF except `output` section.`. It would be great if you detail that a bit further.

It would be better to have the docstring explains what is returned and how the data is being processed (if there is any processing).

Could you attempt to use Sphinx's syntax for the docstring? Something similar to https://github.com/owtf/ptp/blob/develop/ptp/libptp/parser.py#L57 for instance.

framework/db/mapping_manager.py
```diff
@@ -74,6 +74,14 @@ def GetMappingTypes(self):
         """
         return self.mapping_types
 
+    def get_all_mappings(self):
+        """
+        Returns a dictionary with key: owtf_code and value: mapping dictionary.
```
DePierre: Could you try to follow Sphinx's syntax? You can find an example at https://github.com/owtf/ptp/blob/develop/ptp/libptp/parser.py#L57

A suggestion: would you find the following docstring clearer?
```python
def get_all_mappings(self):
    """Create a mapping between OWTF plugins code and OWTF plugins description.

    :return: Mapping dictionary {code: description, code2: description2, ...}
    :rtype: dict
    """
```

@saganshul Thanks for keeping the code base documented btw :) really appreciated!

framework/interface/api_handlers.py
saganshul: @DePierre Changes made. Please review.

DePierre:
@saganshul I'm not very fond of the API docstring. I find it a bit confusing (in addition of some typos).

It is also wrong sometimes. For instance `:raise:` should describe the exceptions that can be raised by the function and `400` is not an exception.

I will be back next week. Until then I can only briefly review from my phone.

framework/interface/api_handlers.py
```diff
@@ -52,39 +52,34 @@ def get(self, plugin_group=None, plugin_type=None, plugin_code=None):
 class PluginNameOutput(custom_handlers.UIRequestHandler):
     SUPPORTED_METHODS = ['GET']
 
-    def get(self, target_id=None, plugin_group=None, plugin_type=None, plugin_code=None):
+    def get(self, target_id=None):
+        """Retrieve plugins details and the corresponsing data from a target scan.
```
viyatb: Typo.

saganshul: @viyatb I am removing present docstring. I think we should use `sphinxcontrib.httpdomain` for API documentation.

viyatb: Ok, is it possible to add the documentation in that format? I'll add that Sphinx plugin in a separate PR.

saganshul: I found this example http://stolarscy.com/dryobates/2016-10/sphinx-rest-api/
I will start adding API documentation, when plugin is integrated with OWTF.

viyatb: ok

saganshul: Done!

viyatb: Rebase please! :)

saganshul: Done! Please review.

PR #875, opened 2017-08-19
Turning OWTF into a Python package

viyatb:
This PR is the first of many steps to refactoring OWTF and make it installable as a Python package.
While most of the changes are self-explanatory, here are a couple of notes on the work:

- Including non-Python files like templates, plugin files, or any directory which is not a Python module is a PITA. I had to add the correct paths to the `MANIFEST.in` file to correct that.
- The new web interface was moved to its separate directory (this was done in an earlier commit). We need to deprecate the old method of rendering templates using Tornado.
-  I completely removed Zest, PlugnHack, WafBypasser and Proxy miner support. We need addons support in OWTF so that optional features can be easily plugged in.
- I have renamed `configuration` to `conf` to separate it from the other config folder.
- The OWTF current install runs a post installation step in `python setup.py install`. I removed the virtualenv setup completely, since now it is the user's job to run `python setup.py install` in a separate virtualenv for maximum compatibility.
- Added Sphinx docstrings to almost every function and module in OWTF
- Convert all function names to snake case.
- All code is now compatible with Python3 and Python2
- Fix tests
- Refactor installation method to install everything to `~/.owtf`.
- Add Debian packaging scripts
- Better Makefile

#### Steps for installation

(I have not yet uploaded the package to PyPi, yet!)

1. Create a new virtualenv, `virtualenv <env>` and activate the environment.
2. Go into OWTF directory and run `python setup.py install` which install OWTF as a package and starts the postsetup install script.
3. To run OWTF, make a new folder for your target engagement, and run OWTF as `python -m owtf`.

After this, OWTF should no longer be responsible for
- running Postgresql on startup (user's job!)
- virtualenv management (users should use it by default for separate projects)

 Let me know if there are any questions.


DePierre:
@viyatb Very happy to see such changes :)

I won't be able to review anything until starting next week though. I hope you won't mind :s Even then it might take time to review so many changes, although most are simply moving things around.

I'm surprised regarding your comment about the PITA of including non-python files. I thought that it was fairly easy but I'm surely remembering wrong. I'll have to check next week see what I can find.

viyatb:
@DePierre no problem, I'll merge it after fixing the tests. This goes on develop branch, so it is not a big deal (the branch is by default unstable and should have breaking changes).
re: adding non-Python files, I had not done this before, so I had to look it up and read the docs lol.

saganshul:
@viyatb `.gitsubmodule` will also need changes:
Also one question, can we move owtf/webui folder out of `owtf` because I don't think python module will require this?

viyatb:
@saganshul re gitmodule: We discussed this on Slack, we'll have to move it to the installation part (download the templates into `~/.owtf`)
 re webui: IMO the web interface is the one of the main parts of the tool, so it should come bundled with the package.

saganshul:
@viyatb re gitmodule: yeah! we can do that. 

So basically, if we have plans to render everything using ReactJS so there are two ways of working in webpack environment. ( I will remove the gulp completely because webpack will be handle development thing.
- **Development**: That will need to install node_modules, webpack etc. These things are only required if we are developing UI. Webpack comes if integrated development environment and working with [react-redux](https://github.com/reactjs/react-redux) is a ++. Only requirement is to render everything using React. https://webpack.github.io/docs/webpack-dev-server.html. Hence using this things development becomes very easier, webpack automatically detects changes using socket and only build for changes. 
Also using https://webpack.github.io/docs/hot-module-replacement.html makes ReactJS development a charm. 

- **Production**: In production environment we actually don't need any of JS from /owtf/webui/src/. We will just need bundle.js, bundle.css to work with. If some user is not developing UI, he/she doesn't need to install any node_modules. Also if we are providing owtf as python package, I think we should just ship **latest** bundle.js and bundle.css(after we move to scss).
 
So moving webui/src/ directory actually make sense. IMO if user is installing/building python module, installer should not build any of **JS** module. Let me know your thoughts on the same.

viyatb:
@saganshul You are right. The only problem there is we would have to add a hack to get the path before the owtf package is installed and add that to the UI config so that it builds the `bundle.js` and adds it to the appropriate path. It would be messy.

saganshul:
@viyatb tell me one thing, after this PR how will user launch the UI?

viyatb:
@saganshul a normal run is sufficient, `python -m owtf`
screen shot 2017-08-19 at 12 45 47

saganshul:
@viyatb One thing we can do we can host bundle.js and bundle.css somewhere and wget them at the time of installation.

viyatb:
Again messy and hacky stuff.

saganshul:
@viyatb Let me think what we can do. I will try things when start rendering things.

viyatb:
@saganshul I believe I can add a flag `dev=False` to the config so that the React bundle and CSS will be loaded from a directory in the pip package instead of generating it everytime with `pip install owtf`

owtf/db/transaction_manager.py
```diff
@@ -170,6 +170,7 @@ def get_first(self, criteria, target_id=None):
         :return:
         :rtype:
         """
+        print criteria, target_id
```
DePierre: Seems like debug output to me.

owtf/db/transaction_manager.py
```diff
@@ -195,7 +196,7 @@ def get_transaction(self, trans):
         :return:
         :rtype:
         """
-        if trans:
+        if (trans and len(trans) > 0):
```
DePierre: Parenthesis are superfluous.

owtf/db/transaction_manager.py
```diff
@@ -217,7 +218,9 @@ def get_transactions(self, transactions):
         """
         owtf_tlist = []
         for transaction_obj in transactions:
-            owtf_tlist.append(self.get_transaction(transaction_obj))
+            print self.get_transaction(transaction_obj)
```
DePierre: Seems like debug output to me.

owtf/db/transaction_manager.py
```diff
@@ -217,7 +218,9 @@ def get_transactions(self, transactions):
         """
         owtf_tlist = []
```
DePierre: Wouldn't it be possible to shorten those lines with:
```python
return [self.get_transaction(transaction) for transaction in transactions if self.get_transaction(transaction) is not None]
```

owtf/db/url_manager.py
```diff
@@ -105,6 +105,19 @@ def add_to_db(self, url, visited, found=None, target_id=None):
...
+    def get_num_urls(self, target_id=None):
+        count = self.db.session.query(models.Url(target_id=target_id)).count()
+        return count
+
+    def add_urls_start(self):
+        self.num_urls_before = self.get_num_urls()
+
+    def add_urls_end(self):
+        num_urls_after = self.get_num_urls()
+        msg = str(num_urls_after - self.num_urls_before)+" URLs have been added and classified"
```
DePierre: Missing space surrounding the `+` operator.

owtf/db/url_manager.py
```diff
@@ -105,6 +105,19 @@ def add_to_db(self, url, visited, found=None, target_id=None):
...
+        logging.info(msg)
```
DePierre: Is this debug output?

owtf/db/url_manager.py
```diff
@@ -105,6 +105,19 @@ def add_to_db(self, url, visited, found=None, target_id=None):
             self.db.session.merge(models.Url(target_id=target_id, url=url, visited=visited, scope=scope))
             self.db.session.commit()
 
+    def get_num_urls(self, target_id=None):
+        count = self.db.session.query(models.Url(target_id=target_id)).count()
+        return count
+
+    def add_urls_start(self):
```
DePierre: Is the name of the function accurate? I am not sure to understand its purpose when I read `add_urls_start` and I see that it stores the number of URLs to a class attribute.

owtf/http/requester.py
```diff
@@ -309,15 +309,15 @@ def request(self, url, method=None, post=None):
...
         self.req_count_total += 1
         try:
             response = self.perform_request(r)
-            self.set_succesful_transaction(raw_request, response)
+            self.set_successful_transaction(raw_request, response)
         except HTTPError as Error:  # page NOT found.
             # Error is really a response for anything other than 200 OK in urllib2 :)
             self.http_transaction.set_transaction(False, raw_request[0], Error)
-        except urlError as Error:  # Connection refused?
+        except URLError as Error:  # Connection refused?
```
DePierre: `Error` should be snake case and not camel case.

owtf/http/transaction.py
```diff
@@ -17,6 +17,7 @@
 except ImportError:
     from httplib import responses as response_messages
 
+from owtf.lib.general import *
```
DePierre: Wildcard import should be avoided at all cost to ease debugging and prevent scope override.

owtf/plugin/plugin_handler.py
```diff
@@ -552,7 +552,7 @@ def process_plugins_for_target_list(self, plugin_group, status, target_list):
                                 {'SomeAborted': False, 'SomeSuccessful': False, 'AllSkipped': True},
                                 {target})
         else:
-            pass
+            exit(0)
```
DePierre: I am not sure about the context. Is the user expecting OWTF to exit in such case? Is it an actual success?

owtf/db/url_manager.py
viyatb: This is not, but the function is not used anyway. I am removing it.

owtf/db/url_manager.py
viyatb: Removing the function.

owtf/db/url_manager.py
viyatb: I ported it from a commit while back because the plugin helper had a method which was calling it. I removed it from there.

owtf/plugin/plugin_handler.py
viyatb: Not needed, was just for debugging. Removing this.

owtf/http/transaction.py
viyatb: +1

owtf/db/transaction_manager.py
viyatb: +1

owtf/db/transaction_manager.py
viyatb: Yes it is.

viyatb:
@DePierre ready to merge? :)

PR #879, opened 2017-08-24
OWTF works work, fixed all refactor mistakes, and improved installation

viyatb:
Fixes
 - plugins not running
 - not recording the plugin output
 - no workers registering the work

Basically there were some functions yet to renamed globally and OWTF silently failed because of this.

This PR also streamlines the installation.


owtf/install/install.py
```diff
@@ -70,158 +62,54 @@ def install_in_directory(directory, command):
...
-    Colorizer.info("[*] Installing required packages for pipsecure")
-    run_command("sudo pip2 install --upgrade pyopenssl ndg-httpsclient pyasn1")
-
-
-def install(cmd_arguments):
-    """Perform installation of OWTF Framework. Wraps around all helper methods made in this module.
+    """Checks if this is a Debian like linux distribution
 
-    :param cmd_arguments:
+    :return: `
```
DePierre: Seems to be a typo in the docstring.

owtf/install/install.py
```diff
@@ -70,158 +62,54 @@ def install_in_directory(directory, command):
...
-                    continue
-            except ValueError:
-                print("")
-                Colorizer.warning("[!] Invalid Number specified")
-                continue
-
-    # Install distro specific dependencies and packages needed for OWTF to work
-    if distro_num != 0:
-        run_command(cp.get(cp.sections()[int(distro_num) - 1], "install"))
+    compatible_value = os.system("which apt-get >> /dev/null 2>&1")
```
DePierre: Likely a stupid remark but this could be abused by a malicious user with `PATH` pointing to a custom directory, in the case OWTF is run with setuid.

Should we use absolute path instead?

owtf/install/install.py
```diff
@@ -230,29 +118,17 @@ def danger(cls, string):
     # Path to custom scripts for tasks such as setting up/ running PostgreSQL db, run arachni, nikto, wapiti etc.
     scripts_path = os.path.join(root_dir, "scripts")
 
-    # OWTF python libraries
-    owtf_pip = os.path.join(root_dir, "install", "owtf.pip")
-
-    # Restricted tools and dictionaries which are distro independent
-    restricted_cfg = os.path.join(root_dir, "install", "distro-independent.cfg")
+    # Copy all necessary directories
+    copy_dirs('conf')
```
DePierre: These directories should not be hardcoded IMO

owtf/interface/templates/manager_interface.html
```diff
@@ -165,6 +165,7 @@ <h4 class="modal-title">Worker Log</h4>
             ifModified: true,
             dataType:'json',
             success:function(data, textStatus, jqXHR){
+                console.log(data);
```
DePierre: Is this necessary in production? Looks like debug output to me.

owtf/interface/templates/worker_interface.html
```diff
@@ -56,6 +56,7 @@
             ifModified: true,
             dataType:'json',
             success:function(obj, textStatus, jqXHR){
+                console.log(obj);
```
DePierre: Same question.

owtf/plugin/plugin_helper.py
```diff
@@ -38,49 +38,65 @@ def __init__(self):
...
-
-    def MultipleReplace(self, Text, ReplaceDict):  # This redundant method is here so that plugins can use it
-        return MultipleReplace(Text, ReplaceDict)
-
-    def CommandTable(self, Command):
+        self.robots_allow_regex = re.compile("Allow: ([^\n  #]+)")
+        self.robots_disallow_regex = re.compile("Disallow: ([^\n #]+)")
+        self.robots_sitemap = re.compile("Sitemap: ([^\n #]+)")
+
+    def multi_replace(self, text, replace_dict):
```
DePierre: Not an appropriate docstring imo. It could be a note instead. The docstring should explain what the function does.

viyatb: I am not refactoring any of this. This should never be done on the server side. Will be ported over to the React app.

owtf/interface/templates/worker_interface.html
viyatb: +1. My bad. I was testing why workers were not being rendered on the UI.

owtf/interface/templates/manager_interface.html
viyatb: Same. Will remove.

owtf/install/install.py
viyatb: These have to be either here or should be fetched from the config file. Basically it copies over `conf`, `dictionaries` and `tools` directories to `~/.owtf` so that the pip package space is not polluted :)

owtf/install/install.py
viyatb: +1 Will fix.

owtf/install/install.py
viyatb: I am going to remove this - this is not used anymore. We support direct Kali installation by `pip install owtf` or Docker on any other platform/distro.

PR #881, opened 2017-09-01
Update proxy.py

Tikam02:
## screenshots
http-https-issue

screenshot from 2017-09-01 17-41-11

screenshot from 2017-09-01 17-59-35

 There was two http connection codes lines, i thought one would be on https.please check it once.because on my system for https and http it runs for it's own connection.
 ## Reviewer
@viyatb

### DIFF
--- a/framework/http/proxy/proxy.py
+++ b/framework/http/proxy/proxy.py
@@ -290,7 +290,7 @@ class ProxyHandler(tornado.web.RequestHandler):
         # HTTPS fail check has to be improvised
         def ssl_fail():
             try:
-                self.request.connection.stream.write(b"HTTP/1.1 200 Connection established\r\n\r\n")
+                self.request.connection.stream.write(b"HTTPS/1.1 200 Connection established\r\n\r\n")
             except tornado.iostream.StreamClosedError:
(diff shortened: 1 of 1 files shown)


DePierre:
@Tikam02 Thanks for taking to time to contribute.

Whether the connection uses TLS or not (HTTP vs. HTTPS) does not change the raw text being exchanged by the two parties. The following HTTP request is send both when you are using following http://google.com and https://google.com:
```http
GET / HTTP/1.1
Host: google.com
```

It is simply how the HTTP protocol is defined (see https://tools.ietf.org/html/rfc2616, section 3.1 HTTP Version):
>  The version of an HTTP message is indicated by an HTTP-Version field
>   in the first line of the message.
> 
>       HTTP-Version   = "HTTP" "/" 1*DIGIT "." 1*DIGIT
It has nothing to do with using TLS vs. not using TLS. It's the layer around that text that will change whether you're using TLS

You could also check https://textslashplain.com/2015/11/19/understanding-connect-tunnels/ which is more specific to the line you changed.

I hope this makes things more clear. Feel free to ask any questions ;)

PR #939, opened 2018-03-09
fix(docker-compose): fix docker-compose ports and install script

fristonio:
- fix ports in `docker-compose.yml`
- fix docker environment check in install script.

## Reviewers
@viyatb 

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ install-docs-requirements:
 install-requirements: install-python-requirements install-node-requirements
 
 react-build:
-	cd owtf/webapp && yarn build
+	cd owtf/webapp && yarn run prod
 
 develop: install-requirements react-build
 
--- a/Makefile
+++ b/Makefile
@@ -44,20 +44,16 @@ install-ui-requirements:
 
 install-docs-requirements:
 	@echo "--> Installing Sphinx dependencies"
+	pip install sphinx sphinx_rtd_theme
 
-install-requirements: install-python-requirements install-node-requirements
+install-requirements: install-python-requirements install-ui-requirements install-docs-requirements
 
-react-build:
-	cd owtf/webapp && yarn run prod
-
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -48,7 +48,7 @@ install-docs-requirements:
 install-requirements: install-python-requirements install-node-requirements
 
 react-build:
-	cd owtf/webapp && yarn build
+	cd owtf/webapp && yarn run prod
```
viyatb: Is it possible to have command `make react-build` take arguments "dev" and "prod" so its flexible to use?

fristonio: I think there is a way we can do this using `make ENV="dev" react-build`.  But I think a cleaner solution would be to have two different make tags for this as we have now, though we can change the tags names to something `make web-dev` and `make web-prod`.

image

viyatb: I think the env approach is more cleaner since we don't need to do code duplication. So, it should look something like,

```
make OWTF_ENV={dev, prod}
```
and use this as `cd owtf/webapp && yarn run $(OWTF_ENV)`. However there should be default value if not env variable is defined.

fristonio: Yeah seems logical to me too. I will push the requested changes. :)

fristonio:
@viyatb Done.

Makefile
```diff
@@ -44,20 +44,16 @@ install-ui-requirements:
...
-
-develop-ui:
-	cd owtf/webapp && yarn run
-
-install-develop-ui: install-node-requirements develop-ui
+web:
+ifdef OWTF_ENV
+	cd owtf/webapp && yarn run ${OWTF_ENV}
+else
+	@echo "--> No environment specified. Usage: make make OWTF_ENV={dev, prod} web"
```
viyatb: Typo in "make".

PR #954, opened 2018-03-21
fix(953): exit cleanly when no postgreSQL db server is running

### LINKED ISSUE #953: OWTF does not exit cleanly when PostgreSQL server is not running.
owtf when run with PostgreSQL not running does not exit cleanly and prints sqlalchemy tracebacks. 

## Expected Behavior
The correct behaviour should make owtf exit cleanly with an error log that postgres server is not running.

## Current Behavior
owtf prints Traceback to the console.

## Possible Solution
owtf should catch the exception raised by SQLAlchemy to print error that postgres server is not running.

1. Stop postgres server running using `sudo service postgres stop`
2. Try running owtf

## Logs
https://pastebin.com/dLJEnWXS

## Your Environment

* Installation method used: Installation script
* Operating System and version (like Kali, Debian, ArchLinux, etc): Debian
* Python version:  3.5.4

fristonio:
- owtf now exits cleanly when no postgreSQL server is running.
- Fixes #953 

## Reviewers
@viyatb 

image

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/owtf/db/database.py
+++ b/owtf/db/database.py
@@ -5,14 +5,15 @@ owtf.db.database
 This file handles all the database transactions.
 """
 import functools
+import sys
 import logging
 
-from sqlalchemy import create_engine, func
+from sqlalchemy import create_engine, exc, func
 from sqlalchemy.orm import Session as _Session
 from sqlalchemy.orm import sessionmaker
 
 from owtf.db.models import Base
-from owtf.settings import DATABASE_IP, DATABASE_NAME, DATABASE_PASS, DATABASE_PORT, DATABASE_USER
+from owtf.settings import DATABASE_IP, DATABASE_NAME, DATABASE_PASS, DATABASE_USER, DATABASE_PORT
 
 
 def get_count(q):
@@ -43,11 +44,16 @@ def flush_transaction(method):
 
 
 def get_db_engine():
-    engine = create_engine(
-        "postgresql+psycopg2://{}:{}@{}/{}".format(DATABASE_USER, DATABASE_PASS, DATABASE_IP, DATABASE_NAME),
-        pool_recycle=120)
-    Base.metadata.create_all(engine)
-    return engine
+    try:
+        engine = create_engine(
+            "postgresql+psycopg2://{}:{}@{}:{}/{}".format(DATABASE_USER, DATABASE_PASS, DATABASE_IP, DATABASE_PORT,
+                                                          DATABASE_NAME),
+            pool_recycle=120)
+        Base.metadata.create_all(engine)
+        return engine
+    except exc.OperationalError as e:
+        logging.error("Could not create database engine - Exception occured\n{}".format(str(e)))
+        sys.exit(1)
 
 
 def get_scoped_session():
--- a/owtf/db/database.py
+++ b/owtf/db/database.py
@@ -52,7 +52,7 @@ def get_db_engine():
         Base.metadata.create_all(engine)
         return engine
     except exc.OperationalError as e:
-        logging.error("Could not create database engine - Exception occured\n{}".format(str(e)))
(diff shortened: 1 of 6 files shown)


owtf/db/database.py
```diff
@@ -6,13 +6,18 @@
...
+from sqlalchemy import create_engine, func
 from sqlalchemy.orm import Session as _Session
 from sqlalchemy.orm import sessionmaker
 
 from owtf.db.models import Base
-from owtf.settings import DATABASE_IP, DATABASE_NAME, DATABASE_PASS, DATABASE_PORT, DATABASE_USER
+from owtf.settings import DATABASE_IP, DATABASE_NAME, DATABASE_PASS, DATABASE_USER
+from owtf.utils.logger import OWTFLogger
+
+logger = OWTFLogger()
```
viyatb: No need to import the logger. Just import the existing instance from `core.py`.

owtf/db/database.py
```diff
@@ -41,11 +46,15 @@ def wrapper(self, *args, **kwargs):
...
 
 def get_db_engine():
-    engine = create_engine(
+    try:
+        engine = create_engine(
             "postgresql+psycopg2://{}:{}@{}/{}".format(DATABASE_USER, DATABASE_PASS, DATABASE_IP, DATABASE_NAME),
-                pool_recycle=120)
-    Base.metadata.create_all(engine)
-    return engine
+            pool_recycle=120)
```
viyatb: Catch the exception like in https://github.com/owtf/owtf/blob/master/owtf/db/database.py#L147.

owtf/db/database.py
fristonio: what do you exactly mean by import existing instance from core.py? because as of now core does not export `logger` also db connection is called even before executing core.

viyatb: Then maybe just `logger = logging.getLogger(__name__)` should do it.

fristonio: But we won't get command line logging formatting that way, is there a problem with the current approach?

owtf/db/database.py
fristonio: Why aren't we using DB_PORT while creating the engine? I think we should use it since it will allow user to explicitly define the port on which postgres is running using `settings.py`.

viyatb: Agreed. Change this to use value from `settings.py`.

owtf/db/database.py
viyatb: I am thinking about unnecessary code duplication here. Can we configure the custom logger in `utils/logger.py` which can be used universally without initializing it everywhere?

fristonio: Let me have a look at it, I will get back to you on this.

fristonio: How about if we do
```python
logger = OWTFLogger()
```
in logger.py and then instead of importing OWTFLogger we can import `logger` then we will only need to enable_logger in the importing model to get what we want.

fristonio:
@viyatb take a look.

owtf/db/database.py
```diff
@@ -43,11 +44,16 @@ def wrapper(self, *args, **kwargs):
...
-    return engine
+    try:
+        engine = create_engine(
+            "postgresql+psycopg2://{}:{}@{}:{}/{}".format(DATABASE_USER, DATABASE_PASS, DATABASE_IP, DATABASE_PORT,
+                                                          DATABASE_NAME),
+            pool_recycle=120)
+        Base.metadata.create_all(engine)
+        return engine
+    except exc.OperationalError as e:
+        logging.error("Could not create database engine - Exception occured\n{}".format(str(e)))
```
viyatb: Make this `logging.error("Could not create database engine - Exception occured\n%s", str(e))` according to https://www.python.org/dev/peps/pep-0282/ best practices (some perf benefits).

PR #1020, opened 2019-04-03
This fixes the following scenario:

diofeher:
When user is installing owtf and doesn't have tornado installed in its own system, when importing __version__ on setup.py, it will try to load OWTFLogger, that uses tornado, then breaking the application.

My fix just bypass that by handling the LoadModule error and installation goes fine.

## How Has This Been Tested?
Tested on OSX 10.12

Before:
Screen Shot 2019-04-02 at 23 24 58

After:
It's working!

## Types of changes


diofeher:
@viyatb

PR #1025, opened 2019-05-23
Targets page, Plugins and Sessions component implemented in evergreen-ui

sharmamohit123:
Implemented the Targets Page, session and plugin component using Evergreen components with minor bug fixes and complete documentation.

## Reviewers

@viyatb @saganshul @7a 

### Targets page
Screenshot from 2019-05-23 05-19-14

### Plugin component
Screenshot from 2019-05-23 05-20-48

### Session component
Screenshot from 2019-05-23 05-21-31


viyatb:
@sharmamohit123 I am curious -why go with the sidebar design in the targets page? We could make that into another modal.

re session modal: Some nits on the whitespace in the columns - can you make them minimal by adjusting the table column width?

sharmamohit123:
@viyatb I tried implementing the previous design but it didn't looked that well with the new targets table. That's why I moved onto the sidebar design (inspired by dropbox). If you want I can use the previous ui and We can use this approach in the reports page. what say?

viyatb:
let's try that - I think that sidebar is more suited for the reports page since we have a lot of information to cram in the page.

sharmamohit123:
sure @viyatb, will do that.

sharmamohit123:
@viyatb this is how the updated targets page looks -
Screenshot from 2019-06-01 01-12-40

Issue #1057, opened 2020-04-23
Manual Installation when setup.py is run


Ashrith-Shetty:
On running python setup.py develop a ModuleNotFoundError is thrown.
We have to manually install the following Modules Tornado, PyYAML, Six.
This issue is present in both manual and docker installation process
 

## Expected Behavior

The installation of all the required dependencies is done automatically.

## Current Behavior

Requires the user to manually install some of the modules.

## Steps to Reproduce (for bugs)

1. python setup.py develop
2. 
3.
4.

## Your Environment

* Installation method used:
* Version/branch used:
* Operating System and version (like Kali, Debian, ArchLinux, etc):
* pip/setuptools version:
* sudo access:

viyatb:
Closing this as stale. The install/Docker setup has changed a lot since this was opened, so this report is no longer actionable as-is. If this still reproduces on current `develop`, please open a fresh issue with OS, Docker/Python versions, and the current logs.

PR #1058, opened 2020-04-23
fix: manual installation of modules

Ashrith-Shetty:
On running setup.py, ModuleNotFoundError was being thrown.
The reason being "from owtf import __version" imports packages from install_requires dependencies which are not yet installed.

## Description

Instead of importing the version number, the code the version will now extract it from init.py

## Related Issue

https://github.com/owtf/owtf/issues/1057

Screenshot from 2020-04-24 04-06-14

## Types of changes

- [ ] Other

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,9 +9,20 @@ except ImportError:
 from setuptools.command.develop import develop
 from setuptools.command.install import install
 
-from owtf import __version__
 
 ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
+
+#Extracts the version details 
+def get_version(init_path):
+    #Read the __init__ file
+    with open(init_path,"r+") as f:
+        init_page = f.read()
+    #Extract the version from the init file
+    for each_line in init_page.splitlines():
+        if each_line.startswith('__version__'):
+            delimiter = '"' if '"' in each_line else "'"
+            return(each_line.split(delimiter)[1])
+
 
 
 def strip_comments(l):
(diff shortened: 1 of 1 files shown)


sharmamohit123:
changes works fine on my system (resolved the issue), @viyatb can you double check and merge this

setup.py
```diff
@@ -9,10 +9,21 @@
...
 from setuptools.command.install import install
 
-from owtf import __version__
 
 ROOT_DIR = os.path.dirname(os.path.realpath(__file__))
 
+#Extracts the version details 
+def get_version(init_path):
+    #Read the __init__ file
+    with open(init_path,"r+") as f:
```
viyatb, line 18: can you format code with `black` (code formatter)?.

setup.py
```diff
@@ -9,10 +9,21 @@
...
+        init_page = f.read()
+    #Extract the version from the init file
+    for each_line in init_page.splitlines():
+        if each_line.startswith('__version__'):
+            delimiter = '"' if '"' in each_line else "'"
+            return(each_line.split(delimiter)[1])
```
viyatb, line 24: Modify the `return(<>)` to be `return <>`

viyatb (changes requested):
A better fix is to remove the circular dependency in the __init__.py file than this though.

setup.py
Ashrith-Shetty: Thank You for your feedback, I will make the appropriate changes.

Issue #946, opened 2018-03-19
Authentication handlers for API and web server


viyatb:
OWTF should provide authentication options for users to log in and securely run tests if OWTF is hosted on a public IP.

For a barebones implementation, we need:

* API handlers for register, login and logout functionality.
* Tokens for OWTF REST API.
* A way for users to create API tokens.
* Users, APITokens models and integration with current DB schema.

Reference:
* https://github.com/jupyter/notebook/blob/master/notebook/auth is a good reference implementation of the functionality (in Tornado).

amanpro30:
Hello @viyatb, I would like to take up this issue.

viyatb:
Closing this as stale to keep the roadmap focused. The idea may still be worth revisiting, but this issue has not had recent ownership or a concrete plan. A fresh proposal against current `develop` would be the better place to pick it up.

PR #1105, opened 2021-03-12
Adding Authentication handlers for API and Web Server

amanpro30:
## Description
Created the API for login, register, and logout functionality. The register API accepts email, password, confirm_password, and name of the new user, and hence I updated the User model to accommodate the same. The login returns the JWT Token on successful login and sets the cookie for the user. I also created a model for API tokens having user_id and api_key fields and API endpoint for creating API tokens which generate an API token for the authenticated user.

Tried implementing the following:
- [x]  API handlers for the register, login, and logout functionality.
- [x] Tokens for OWTF REST API (ideally JWT tokens)
- [x] A way for users to create API tokens.
- [x] Users, APITokens models, and integration with current DB schema.

## Related Issue
Fix for #946 

## Reviewers
@viyatb @sharmamohit123 

## How Has This Been Tested?
I have tested all the APIs using PostMan

image
image
image

## Types of changes
- [ ] Other

### Checklist:

### DIFF
--- a/owtf/api/handlers/api_token.py
+++ b/owtf/api/handlers/api_token.py
@@ -8,10 +8,8 @@ from owtf.api.handlers.jwtauth import jwtauth
 from uuid import uuid4
 from owtf.models.api_token import ApiToken
 import jwt
-from owtf.settings import JWT_SECRET_KEY
+from owtf.settings import JWT_SECRET_KEY, JWT_OPTIONS
 from owtf.lib.exceptions import APIError
-
-options = {"verify_signature": True, "verify_exp": True, "verify_nbf": False, "verify_iat": True, "verify_aud": False}
 
 
 @jwtauth
@@ -52,7 +50,7 @@ class ApiTokenGenerateHandler(APIRequestHandler):
         api_key = str(uuid4())
         try:
             token = self.request.headers.get("Authorization").split()[1]
-            payload = jwt.decode(token, JWT_SECRET_KEY, options=options)
+            payload = jwt.decode(token, JWT_SECRET_KEY, options=JWT_OPTIONS)
             user_id = payload.get("user_id", None)
             if not user_id:
                 raise APIError(400, "Invalid User Id")
--- a/owtf/api/handlers/api_token.py
+++ b/owtf/api/handlers/api_token.py
@@ -1,0 +1,42 @@
+"""
+owtf.api.handlers.auth
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+"""
+from owtf.api.handlers.base import APIRequestHandler
+from owtf.api.handlers.jwtauth import jwtauth
+from uuid import uuid4
+from owtf.models.api_token import ApiToken
+import jwt
+secret_key = "my_secret_key"
+
+options = {
+    'verify_signature': True,
+    'verify_exp': True,
+    'verify_nbf': False,
+    'verify_iat': True,
+    'verify_aud': False
+}
+
+@jwtauth
+class ApiTokenGenerateHandler(APIRequestHandler):
+    """
+        doc
+    """
+    SUPPORTED_METHODS = ["GET"]
+
+    def get(self):
+        api_key = str(uuid4())
+        token = self.request.headers.get('Authorization').split()[1]
+        payload = jwt.decode(
+            token,
+            secret_key,
+            options=options
+        )
+        user_id = payload['user_id']
+        ApiToken.add_api_token(self.session, api_key, user_id)
+        data = {
+            "status": "ok",
+            "api_key": api_key
+        }
+        return self.success(data)
--- a/owtf/api/handlers/api_token.py
+++ b/owtf/api/handlers/api_token.py
@@ -1,5 +1,5 @@
 """
-owtf.api.handlers.auth
+owtf.api.handlers.api_token
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 """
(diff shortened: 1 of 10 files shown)


owtf/api/handlers/api_token.py
```diff
@@ -0,0 +1,42 @@
...
+owtf.api.handlers.auth
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+"""
+from owtf.api.handlers.base import APIRequestHandler
+from owtf.api.handlers.jwtauth import jwtauth
+from uuid import uuid4
+from owtf.models.api_token import ApiToken
+import jwt
+secret_key = "my_secret_key"
```
viyatb: can this be defined in settings and imported here? Its a bad practice to hard-code a secret in code.

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+from datetime import datetime, timedelta
+import tornado
+import bcrypt
+import json
+import base64
+import uuid
+import jwt
+import re
+
+JWT_SECRET = 'my_secret_key'
```
viyatb: ditto for lines 18-20 - all of this should be in settings

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+                "status": "success",
+                "data": {
+                    "status": "fail",
+                    "message": "Invalid login credentials"
+                }
+            }
+
+        """
+        body_data = json.loads(self.request.body.decode('utf-8'))
+        email = body_data['email']
```
viyatb: in order for this to work correctly, you should catch exceptions gracefully,

```
email = body_data.get('email', None)
if email is not None:
```

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+        password = body_data['password']
```
viyatb: same here

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+
+        else:
+            error_msg = "Invalid login credentials"
+            data = {
+                "status": "fail",
+                "message": error_msg
+            }
+            self.success(data)
+
+    def set_current_user(self, user):
```
viyatb: instead of setting a cookie, is it possible to use `Authorization: Bearer {}` everywhere - this basically mitigates any CSRF attack at all and makes code simpler by not having to manage cookies.

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+                "status": "success",
+                "data": {
+                    "status": "fail",
+                    "message": "Email already exists"
+                }
+            }
+
+        """
+        body_data = json.loads(self.request.body.decode('utf-8'))
+        name = body_data['name']
```
viyatb: catch exceptions - ditto

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+        email = body_data['email']
+        password = body_data['password']
+        confirm_password = body_data['confirm_password']
+
+        already_taken = User.find_by_email(self.session, email)
+        reg = "^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)(?=.*[@$!%*#?&])[A-Za-z\d@$!#%*?&]{8,20}$"
```
viyatb: do you have a test that this regex works as expected?

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+            data["message"] = "Password doesn't match"
+            self.success(data)
+        elif not mat:
+            data["message"] = "Choose a strong password"
+            self.success(data)
+        elif already_taken:
+            data["message"] = "Email already exists"
+            self.success(data)
+        else:
+            hashed_pass = bcrypt.hashpw(password.encode('utf-8'), bcrypt.gensalt(8))
```
viyatb: why only a 8 byte salt? the correct usage is to use the default `bcrypt.gensalt()`

owtf/api/handlers/auth.py
```diff
@@ -0,0 +1,230 @@
...
+
+            {
+                "status": "success",
+                "data": {
+                    "status": "ok"
+                }
+            }
+
+        """
+        self.clear_cookie("user")
```
viyatb: if we do not manage cookies, the only thing we need to do on a log out is invalidate the JWT of the user - its a much simpler state management than cookies.

owtf/api/handlers/jwtauth.py
```diff
@@ -0,0 +1,64 @@
+# """
+#     JSON Web Token auth for Tornado
+# """
+
+import jwt
+secret_key = "my_secret_key"
```
viyatb: ditto - you have multiple places where this "key" is defined, it should be in a settings file

viyatb (changes requested):
finished a first pass and suggested some changes! nice work :)

owtf/models/user.py
```diff
@@ -4,20 +4,40 @@
...
-
+import uuid
 
 class User(Model):
     __tablename__ = "users"
 
     id = Column(Integer, primary_key=True, autoincrement=True)
-    name = Column(Unicode(255), unique=True)
+    name = Column(Unicode(255))
     cookie_id = Column(Unicode(255), default=None, nullable=False, unique=True)
```
viyatb: can we remove the `cookie_id`?

owtf/models/user.py
```diff
@@ -4,20 +4,40 @@
...
+        return session.query(cls).filter_by(email=email).all()
+
+    @classmethod
+    def add_user(cls, session, user):
+        """Adds an user to the DB"""
+        new_user = cls(
+            name = user["name"],
+            email = user["email"],
+            password = user["password"].decode('utf-8'),
+            cookie_id = str(uuid.uuid4())
```
viyatb: remove cookie_id

owtf/settings.py
```diff
@@ -215,3 +201,8 @@
         exec(settings, globals(), locals())
 except FileNotFoundError:
     pass
+
+# JWT
+JWT_SECRET_KEY = "epb$9n04zst7l%wr9iw$cugk2*_^_aud5=snyn1^k*b*)ab@^i"
```
viyatb: add a default one -> something like "changeme" and add a comment to add a new secret key after installation?

viyatb (changes requested):
minor nits, otherwise it looks good!

owtf/settings.py
amanpro30: image
After installation, showing this message to change the default JWT_SECRET_KEY in settings.py

owtf/models/user.py
amanpro30: Removed.

owtf/models/user.py
amanpro30: Yes, there is no need for this now. Updated it in the latest commit.

owtf/api/handlers/api_token.py
amanpro30: image
Added it in settings.py itself and imported it in every file wherever needed.

owtf/api/handlers/auth.py
amanpro30: Changed.

owtf/api/handlers/auth.py
amanpro30: image
Handled it at all the required locations.

owtf/api/handlers/auth.py
amanpro30: Modified.

owtf/api/handlers/auth.py
amanpro30: Removed this part as it is no longer required as I had planned to store JWT token in the client-side and sending it in Authorization header as suggested.

owtf/api/handlers/auth.py
amanpro30: Handling it now.

owtf/api/handlers/auth.py
amanpro30: Corrected.

owtf/api/handlers/jwtauth.py
amanpro30: Noted and changed.

owtf/api/handlers/auth.py
amanpro30: Added UserLoginToken model for storing the JWT token of the user which gets revoked when the user logs out.

owtf/api/handlers/auth.py
amanpro30: Moved this password check regex in settings.py with other such regexes. Also, added another regex for email validation.
image

Password validation checks whether the password is strong enough by checking:
- small Letters[a-z]
- Capital letters[A-Z]
- digit[0-9]
- Special Characters(Except Space)
- Atleast 8 chars and almost 20 chars.

Shown below are a few examples I run against this regex at this platform https://regex101.com/.
image

Similar test results for email validation regex:
image

viyatb (approved):
looks good! minor nit: you have `options = {"verify_signature": True, "verify_exp": True, "verify_nbf": False, "verify_iat": True, "verify_aud": False}` in multiple files - can you add it to one place and import it from there?

PR #1150, opened 2021-06-27
Migrating from Travis CI to GitHub Actions.

EXTREMOPHILARUM:
Travis CI is going to end its free program for open-source organizations.
Github actions provide free executions for public repositories.
This PR provides the necessary configuration files to do so. 

## Motivation and Context

To maintain our current CI pipeline even after the Travis CI program ends we need to migrate to Github actions.

## Reviewers

@viyatb @sharmamohit123 

## How Has This Been Tested?

I have tested this on my own fork extensively and have arrived at the current configuration.

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,7 +3,7 @@ name: OWTF CI
 on:
   push:
     branches: [master, develop]
-  pull_request:
+  pull_request_target:
     branches: [master, develop]
 
   workflow_dispatch:
@@ -33,7 +33,6 @@ jobs:
           --health-retries 5
     steps:
       - uses: actions/checkout@v2
-
 
       - uses: actions/cache@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -0,0 +1,87 @@
...
+    strategy:
+      matrix:
+        python-version: [3.9]
+        node-version: [15.0]
+    services:
+      db:
+        image: postgres
+        env:
+          POSTGRES_USER: owtf_db_user
+          POSTGRES_PASSWORD: jgZKW33Q+HZk8rqylZxaPg1lbuNGHJhgzsq3gBKV32g=
```
viyatb, line 24: I know this is for tests, but we can store this secret as a Github Actions secret (so that it is not checked in git)

setup.py
```diff
@@ -90,7 +90,7 @@ def run(self):
     zip_safe=False,
     platforms="any",
     install_requires=reqs("base.txt") + reqs("docs.txt"),
-    test_requires=reqs("test.txt"),
+    # test_requires=reqs("test.txt"),
```
viyatb: if we don't need this, then we should remove the dead code.

viyatb (changes requested):
Thanks for working on this - I pointed out minor nits

.github/workflows/main.yml
EXTREMOPHILARUM: Sure, can you create the secrets and share them so that I can add them here.

setup.py
EXTREMOPHILARUM: Sure will remove this.

viyatb:
as cleanup, can you remove the existing `.travis.yml` file?

PR #1152, opened 2021-07-10
Security Enhancements preventing unauthenticated access

amanpro30:
## Description
Added some additional security headers to APIRequestHandler and protected all the backend API routes from unauthenticated access.

## Reviewers
@7a @viyatb @sharmamohit123 @saganshul @Ashrith-Shetty 

## Types of changes

- [ ] Other

### Checklist:


owtf/api/handlers/base.py
```diff
@@ -45,6 +45,14 @@ class BaseRequestHandler(RequestHandler):
     def set_default_headers(self):
         self.add_header("X-OWTF-Version", __version__)
         self.add_header("X-Frame-Options", "SAMEORIGIN")
+        self.add_header("X-XSS-Protection", "1; mode=block")
+        self.add_header("X-Content-Type-Options", "nosniff")
+        self.add_header("Strict-Transport-Security", "max-age=31536000; includeSubDomains")
```
viyatb: I think this will break functionality with https?

amanpro30: Removed.

owtf/webapp/src/containers/WorklistPage/api.js
```diff
@@ -1,20 +1,22 @@
 import Request from "../../utils/request";
 import { API_BASE_URL } from "../../utils/constants";
 
+const options = {
+  headers: {
+    "Content-Type": "application/x-www-form-urlencoded;charset=UTF-8",
+    Authorization: "Bearer " + localStorage.getItem("token")
```
viyatb, line 7: doesn't this have to be in quotes? `"Authorization"`

viyatb (changes requested):

owtf/webapp/src/containers/WorklistPage/api.js
amanpro30: This is correct and is working fine when tested. The one with quotes is not working and also when I use auto-formatting it is removing the quotes.

amanpro30: Hi @viyatb,
I searched this on google and found this link: https://stackoverflow.com/questions/60790742/vs-code-node-js-prettier-auto-formatting-messing-up-the-dictionary-key
JS object properties **do not need quotes** except when the property name needs to contain normally invalid characters, like spaces, or dashes, etc.

viyatb: weird 👍

PR #1161, opened 2021-08-12
Allowing login by username also and small UI changes

amanpro30:
## Description
Improved the login module by allowing login and forgot password by username instead of just email as earlier. Also, made some minor UI changes for logout button.

## Reviewers
@7a @viyatb @sharmamohit123 @saganshul 

image
image

## Types of changes

- [ ] Other

### Checklist:


owtf/models/user.py
```diff
@@ -13,7 +13,7 @@ class User(Model):
     __tablename__ = "users"
 
     id = Column(Integer, primary_key=True, autoincrement=True)
-    name = Column(Unicode(255))
+    name = Column(Unicode(255), nullable=False, unique=True)
```
viyatb, line 16: are you sure you want this set to `unique=True`? People have same names all the time xD

viyatb (changes requested):

owtf/models/user.py
amanpro30: name here refers to the unique username/ unique handle. It is not like the first or last name of the user. It is important to keep it unique to allow login by both username or email.

viyatb: ah perfect

PR #1233, opened 2023-07-30
Gobuster dir, bbot sub-domain and AJAX

ph1ne4s:
Added new tools in Application discovery and subdomain takeover.
Resolved issue #1235 .
Addes AJAX vuln. resources.

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] New feature (non-breaking change which adds functionality)
- [ ] Other

### Checklist:

- [ x] My code follows the code style (modified PEP8) of this project.


owtf/managers/resource.py
ph1ne4s, line 1: Resolves #1235

owtf/managers/resource.py
```diff
@@ -105,7 +105,7 @@ def get_resources_from_file(resource_file):
         resource_file, "r"
     ).read().splitlines()  # To remove stupid '\n' at the end
     for line in config_file:
-        if "#" == line[0]:
+        if line.startswith("#"):
```
viyatb, line 108: `line` here is a list not a string

viyatb (changes requested):
minor nit, everything else looks good

owtf/managers/resource.py
ph1ne4s: Actually `line` is a string inside `config_file` which is a list, and so, `startwith` function is running fine without error.
str

ph1ne4s: Actually `line` is a string inside `config_file` which is a list, and so, `startwith` function is running fine without error.
str

viyatb (approved):

Issue #1241, opened 2024-02-25
Build Failure with python versions > 3.9


Pratham1812:
## Expected Behavior

After following the instructions given in the `README.md` to setup the project on kali linux. I expected it to build without any bugs.

## Current Behavior

I constantly faced deprecation warnings as some of the dependencies needed were not compatible with python3.11.8 (version I tested on)
The dependencies which are failing the build when using the versions as specified in `base.txt` - `psycopg2-binary`,`pycurl`,`PyYAML
`,`tornado`
![image](https://github.com/owtf/owtf/assets/32198580/0bf78401-40ce-4f18-9c51-a02bd7110100)
This was due to `PyYAML`

![image](https://github.com/owtf/owtf/assets/32198580/ee3cad83-ffe4-48b0-8969-d57a912b5c3d)
This was due to `pycurl` and similarly for other packages

## Possible Solution

One can possibly migrate the versions of the listed packages to there appropriate latest versions such that the application does not break
## Steps to Reproduce (for bugs)

1. Follow the instructions as listed in the `README.md` file
2. Run `python3 setup.py develop`

## Context

I was trying to do non docker installation.

## Your Environment

* Installation method used:
* Version/branch used: develop
* Operating System and version (like Kali, Debian, ArchLinux, etc): Kali
* pip/setuptools version: 24.0
* sudo access: Yes

Pratham1812:
@viyatb @7a any help will be appreciated.

Pratham1812:
![image](https://github.com/owtf/owtf/assets/32198580/524b9510-9310-4619-9856-1652a055059f)
Hey I tried fixing the issue with worker logs, right now I am not able to find any need for `owtf/files/routes.py`. Could you please check.
@7a @viyatb

PR #1242, opened 2024-02-28
resolves issue #1241

Pratham1812:
Hey I have tried fixing #1241 the project now functions  with my python version `3.11.8`.I have resolved the issue of deprecated dependencies. I have also removed one of the dependency `cookies` since it was very outdated, instead i have implemented a one liner approach for the use case of this dependency.I have also fixed the login error which several users were facing :)

## Motivation and Context

The change tries to run owtf on systems with python versions > 3.9

## Reviewers

@viyatb @7a 

image
image

Application running without docker with python versions 3.11.8

## Types of changes

- [ ] Other

### Checklist:


owtf/transactions/base.py
```diff
@@ -174,8 +172,8 @@ def get_session_tokens(self):
         cookies = []
         try:  # parsing may sometimes fail
             for cookie in self.cookies_list:
-                cookies.append(Cookie.from_string(cookie).to_dict())
```
viyatb, line 177: can you use `http.cookies` instead of `cookies` package? The reason I prefer to use a std library is that it will handle edge cases better than normal split and get the value.

owtf/models/user_login_token.py
```diff
@@ -28,7 +28,7 @@ def find_by_userid_and_token(cls, session, user_id, token):
     @classmethod
     def add_user_login_token(cls, session, token, user_id):
         """Adds an user_login_token to the DB"""
-        new_token = cls(user_id=user_id, token=token.decode("ascii"))
```
viyatb: I would keep the `.decode` everywhere to handle weird characters - its better to be explicit than implicit :)

viyatb (changes requested):
added some comments

Pratham1812:
image
Hey I tried fixing the issue with worker logs, right now I am not able to find any need for `owtf/files/routes.py`. Could you please check.
@7a @viyatb

owtf/transactions/base.py
```diff
@@ -174,8 +172,9 @@ def get_session_tokens(self):
...
-        except InvalidCookieError:
+                simpleCookie = SimpleCookie(cookie).items()
+                cookies.append([{'name':key,'value':morsel.value} for key,morsel in simpleCookie][0])
```
viyatb: did you test this? it seems wrong

```

In [1]: cookie = 'MUID=12345678910111213141516; domain=.bing.com; expires=Sun, 20-Sep-2020 16:55:46 GMT; path=/;, MUIDB=16151413121110987654321; path=/; httponly; expires=Sun, 20-Sep
   ...: -2020 16:55:46 GMT'

In [2]: from http.cookies import SimpleCookie

In [3]: SimpleCookie(cookie)
Out[3]: <SimpleCookie: >

In [4]: SimpleCookie(cookie).__dict__
Out[4]: {}
```

viyatb: lets replace this with a function from Django source code,

```
def parse_cookie(cookie: str) -> Dict[str, str]:
    """Parse a ``Cookie`` HTTP header into a dict of name/value pairs.

    This function attempts to mimic browser cookie parsing behavior;
    it specifically does not follow any of the cookie-related RFCs
    (because browsers don't either).

    The algorithm used is identical to that used by Django version 1.9.10.

    .. versionadded:: 4.4.2
    """
    cookiedict = {}
    for chunk in cookie.split(str(";")):
        if str("=") in chunk:
            key, val = chunk.split(str("="), 1)
        else:
            # Assume an empty name per
            # https://bugzilla.mozilla.org/show_bug.cgi?id=169091
            key, val = str(""), chunk
        key, val = key.strip(), val.strip()
        if key or val:
            # unquote using Python's algorithm.
            cookiedict[key] = _unquote_cookie(val)
    return cookiedict
```

owtf/api/handlers/jwtauth.py
```diff
@@ -4,7 +4,8 @@
 from sqlalchemy.sql.functions import user
 from owtf.models.user_login_token import UserLoginToken
 import jwt
-from owtf.settings import JWT_SECRET_KEY, JWT_OPTIONS, JWT_ALGORITHM
+from owtf.settings import JWT_SECRET_KEY, JWT_OPTIONS,JWT_ALGORITHM
```
viyatb: the spaces are off here

viyatb (changes requested):
@Pratham1812 did you test the `SimpleCookie` implementation to see if it works?

Pratham1812:
I have updated the way cookies ae handled also did the other changes. Could you please have a look? @viyatb

viyatb (approved):

viyatb:
do you need this if you have the other diff?

PR #1255, opened 2024-03-20
Add password messages

### LINKED ISSUE #1170: Password is marked as "insecure" even though it fits all the requirements.
## Expected Behavior
Upon signing up on the OWTF web interface, the password used (32 characters generated from a password manager with all possible characters) should be marked as secure and should allow me to create an account.

## Current Behavior
I am essentially locked out of the site since any and all passwords (even if they logically fit the requirements to be deemed "secure") are denied and marked as "not secure".
For context, the generated passwords look like this: **v5h^L7Qt%%^6x%ik62z7U$66uX62#&ys7^&@b!**

## Possible Solution
Unknown; I've tried manually going into the postgres database and creating an entry manually to bypass the register page. However, I'm unaware of the exact hashing algorithm used, and the "reset password" functionality probably would result in the same "insecure password" issue.

**Ignore the "invalid email" thing at the start, I changed it later on in the screenshots and the email issue appears unrelated to the password issue**
1. Login Screen without password
2. Login Screen with password
3. Proof that 'confirm password' is not the issue
4. Password Error
5. Denied

## Logs
**The 'Invalid Salt' error might be due to my attempt at manually going into the PSQL database and adding an entry for myself manually. Though I'm not sure if it's on my end or if it is on the server code end.**
[ui_server.log](https://pastebin.com/vWrjTL3M)

## Your Environment
* Installation method used: Docker-Compose
* Version/branch used: 2.5.0
* Operating System and version (like Kali, Debian, ArchLinux, etc): Kali docker container inside Parrot OS.
* pip/setuptools version: Python3.9; Pip 20.3.4
* sudo access: Am able to use root.

AleksaMCode:
Adds additional messages when dealing with passwords to enhance the UX.

## Description

When entering long passwords, users were confronted with a message "_Must have capital, small, number & special chars_." This would cause confusion with users as the password contained all the needed characters. The long password were rejected because the Regex expression requires passwords to be of a certain length, between 8 and 15 characters in length exactly. However, the end user doesn't know that there are password length limitations and is left to wonder why the password isn't accepted.

With these changes, when registering or changing their password, users will be meet with messages "_Password is too short_" or "_Password is too long_" when the password isn't of the appropriate length.

## Related Issue

Closes #1170 

## How Has This Been Tested?

Tested manually

## Types of changes

- [ ] Other

### Checklist:

### DIFF
--- a/owtf/webapp/src/containers/NewPasswordPage/index.tsx
+++ b/owtf/webapp/src/containers/NewPasswordPage/index.tsx
@@ -53,9 +53,11 @@ export class NewPasswordPage extends React.Component<propsType, stateType>  {
       typeof this.state.newPassword !== "undefined"
     ) {
       if (this.state.newPassword.length < 8) {
-        error["newPassword"] = "Password is too short";
+        formIsValid = false;
+        errors["newPassword"] = "Password is too short";
       } else if (this.state.newPassword.length > 15) {
-        error["newPassword"] = "Password is too long";
+        formIsValid = false;
+        errors["newPassword"] = "Password is too long";
       } else if (
         !this.state.newPassword.match(
           /^(?=.*\d)(?=.*[a-z])(?=.*[A-Z])(?=.*[^a-zA-Z0-9])(?!.*\s).{8,15}$/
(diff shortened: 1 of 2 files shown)


viyatb:
@AleksaMCode this will also need a check on the backend side - simply frontend check won't work unless 
https://github.com/owtf/owtf/blob/develop/owtf/models/user.py#L19C1-L20C1 is changed.

AleksaMCode:
@viyatb I'm not sure why this was closed as it helps close an open issue, and it also improves already existing frontend code. Yes, ideally, you would have both frontend and backend password validation, but that issue can be addressed in another PR.

Issue #1259, opened 2024-03-25
Email validation bug


AleksaMCode:
Due to the utilization of `^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$` regex a good deal of incorrect emails are permitted to be used while some other, valid, email addresses are marked as not valid. While this regex does catch most of the valid emails there are still some edge cases which don't get caught. 

## Expected Behavior

These emails shouldn't be marked as not valid: `myemail+owtf@domain.com`, `mike.o'connor@domain.com`
These emails should be marked as not valid: `myemail@domain...com`, `my....email@domain.com`, `myemail@.domain.com`

## Current Behavior

These emails are marked as not valid: `myemail+owtf@domain.com`, `mike.o'connor@domain.com`
These emails are marked as valid: `myemail@domain...com`, `my....email@domain.com`, `myemail@.domain.com`

## Possible Solution

The Regex used for email validation needs to be updated.

## Steps to Reproduce (for bugs)

1. Go to the `Sign Up` page.
2. Enter your email.

skools-here:
The current regex:

^[a-zA-Z0-9._-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]{2,4}$


rejects valid emails like myemail+owtf@domain.com, mike.o'connor@domain.com and accepts invalid ones like myemail@domain...com, myemail@.domain.com.

A more robust alternative is:

^(?!.*\.\.)[a-zA-Z0-9._%+!'’-]+@[a-zA-Z0-9-]+(\.[a-zA-Z0-9-]+)*\.[a-zA-Z]{2,}$


This fixes the above issues and supports modern TLDs.

viyatb:
fixed in [Improve email validation and coverage #1324](https://github.com/owtf/owtf/pull/1324)

PR #1324, opened 2025-11-04
Improve email validation and coverage

viyatb:
## Summary
- replace the duplicated email validation regex with a shared utility that accepts expected characters and blocks invalid dot sequences
- reuse the new validation helper in the signup and forgot password flows and expand their tests to cover valid and invalid examples

## Testing
- CI=1 npm test -- SignupPage *(fails: jest cannot resolve the built-in module alias `node:stream` in this environment)*


viyatb:
fixes https://github.com/owtf/owtf/issues/1259

PR #1334, opened 2025-11-25
Bugfix: Display Actual Running State and Start Time for Workers

### LINKED ISSUE #1330: Worker status (“Running”/“Not Running”) and start time (“N/A”) not updating in UI when plugins running
### **Description**

When any plugin or job is executed, the Worker Panel and the detailed Workers page do not display which worker is running a job. Specifically:

- All workers always show **“Not Running”** in the Worker Panel, even during active scan progress.
- In the detailed worker view, **“Start Time”** always appears as **“N/A”** for all workers, even after starting or completing jobs.
- This makes it impossible for users to see which worker is handling tasks or when those tasks started.

---

### **Steps to Reproduce**

1. Start a plugin or full scan on a target.
2. Observe the Worker Panel: all workers display **“Not Running”** throughout the scan, despite the progress bar updating.
3. Open the detailed worker status/logs page: each worker’s **“Start Time”** shows **“N/A”**, even after jobs are started or completed.
4. Check the logs: workers are performing tasks, but the UI does not update their running status or start time.

---

### **Expected Behavior**

- At least one worker should show **“Running”** and reflect the task assigned, while idle workers show **“Not Running”**.
- **“Start Time”** should display the timestamp when the job began, and only show **“N/A”** when the worker has never been used.

---

### **Screenshots**

Image

Image

piyush140104:
## Description
This PR resolves the issue where the log component displayed logs incorrectly for workers shown as "Not Running" and failed to show the correct start times.

- Updated `owtf/api/main.py`: Added a periodic callback to manage worker status updates and improved error handling.
- Updated `owtf/managers/worker.py`: Improved worker status and start time reporting using `datetime`.
- Updated `WorkerPanel.tsx`: Improve conditional rendering of worker start time (“N/A” if undefined) .

## Related Issue
Fixes #1330 

## Motivation and Context
Accurate worker status and start times are essential for understanding scan progress and debugging. This fix helps users track jobs and analyze scan results more reliably.

## Reviewers
@viyatb 
@7a 

## How Has This Been Tested?
- Manual testing with various plugin and scan jobs.
- Verified workers display “Running” status and correct start times.
- Checked log component updates in real time for active and idle workers.

## Screenshots :

**Before fix:**  

Screenshot 2025-11-25 at 3 46 09 PM
Screenshot 2025-11-25 at 3 37 56 PM
Screenshot 2025-11-25 at 3 38 23 PM

**After fix:**  

Screenshot 2025-11-25 at 3 45 26 PM
Screenshot 2025-11-25 at 3 45 44 PM
Screenshot 2025-11-25 at 3 46 30 PM

## Types of changes
- [ ] Other

### Checklist:


piyush140104:
Hey @viyatb if you have a moment, can you please review this ?

viyatb:
@piyush140104 I see a potential concurrency regression in this PR.

`owtf/api/main.py` now starts a `PeriodicCallback(worker_manager.manage_workers, 2000)` inside the API process.  
But OWTF already runs worker scheduling in the main process via `core.poll_workers()` (`owtf/core.py`).

Because `worker_manager` is a process-local singleton (`worker_manager = WorkerManager()` in `owtf/managers/worker.py`), this effectively creates a second independent scheduler. That can race on work claiming, especially since `get_work_for_target()` does `query.first()` + `delete()` + `commit()` without a lock/transactional claim (`owtf/managers/worklist.py`).

Could we keep worker scheduling in a single process (existing `core.poll_workers()` path), or add proper DB-level locking/atomic claim semantics before introducing another scheduler loop?

piyush140104:
Hey @viyatb,
The latest change (with_for_update(skip_locked=True)) to the database query fix your above concern. Before this, when two scheduler loops ran at the same time, both would read the same row from the database.
With this change, the moment one process picks a row, the database puts a lock on it. If a second process comes along and tries to grab the same row, it sees the lock and just skips over it and picks the next available row instead. So now two processes can never claim the same work item. Each one always gets its own unique task, and the double dispatch bug goes away entirely.


I tried to explore moving scheduler to common place but that was not working. hence, adding atomic looking at db level to fix this issue.

PR #1370, opened 2026-02-16
Fix missing return in auth validation blocks

### LINKED ISSUE #1360: Auth handlers missing return after self.success() causes RuntimeError on invalid input
**Labels:** `bug`, `priority: high`

### Description

In `owtf/api/handlers/auth.py`, the `LoginHandler.post()` and `RegisterHandler.post()` methods call `self.success(err)` when validation fails, but never `return` after it. Since `self.success()` calls Tornado's `self.finish()`, execution continues into subsequent code. If multiple fields are missing, `finish()` gets called more than once, which raises a `RuntimeError: Cannot write() after finish()`.

For example, in `LoginHandler.post()` (lines 97–103):

```python
if not email_or_username:
    err = {"status": "fail", "message": "Missing email or username value"}
    self.success(err)       # no return — execution keeps going
if not password:
    err = {"status": "fail", "message": "Missing password value"}
    self.success(err)       # finish() called again → RuntimeError
user = User.find_by_email(self.session, email_or_username)  # runs with None
```

Same pattern in `RegisterHandler.post()` (lines 186–197) — four consecutive validation checks, none of which return.

### Steps to Reproduce

1. Send a POST to `/api/v1/login` with an empty body (no `emailOrUsername` or `password`)
2. Server hits `self.success()` twice → Tornado raises `RuntimeError`
3. Alternatively, POST to `/api/v1/register` with only `username` set — the code falls through to `re.search(is_password_valid_regex, password)` where `password` is `None`, causing a `TypeError`

### Expected Behavior

Return a single error response and stop processing.

### Actual Behavior

Crashes with `RuntimeError` or `TypeError`. This is likely the root cause behind issue #1239 ("Login have Error: Internal Server Error").

### Fix

Add `return` after every `self.success(err)` call in validation blocks throughout `auth.py`

---

avinxshKD:
Fixes #1360

Added return statements after self.success() in validation blocks where execution continues. Only added to independent if blocks that fall through, not if/elif/else chains.

### DIFF
--- a/owtf/api/handlers/auth.py
+++ b/owtf/api/handlers/auth.py
@@ -97,9 +97,11 @@ class LogInHandler(APIRequestHandler):
         if not email_or_username:
             err = {"status": "fail", "message": "Missing email or username value"}
             self.success(err)
+            return
         if not password:
             err = {"status": "fail", "message": "Missing password value"}
             self.success(err)
+            return
         user = User.find_by_email(self.session, email_or_username)
         if user is None:
             user = User.find_by_name(self.session, email_or_username)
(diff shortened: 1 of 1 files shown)


avinxshKD:
@viyatb pls look into it, small fix

viyatb:
Closing this as part of the current PR backlog cleanup. If the change is still needed, please feel free to reopen it or send an updated PR against the current branch.

PR #1405, opened 2026-03-31
fix: use logging.warning() instead of logging.log() in paths_exist()

### LINKED ISSUE #1404: Bug: logging.log() called with string instead of numeric level in owtf/utils/strings.py
## Describe the bug

In `owtf/utils/strings.py`, `paths_exist()` calls `logging.log()` with a format string as the first argument instead of a numeric log level. `logging.log(level, msg, ...)` requires an integer for the first parameter (e.g. `logging.WARNING`); passing a string raises a `TypeError` at runtime whenever any path in the list does not exist.

```python
# owtf/utils/strings.py, line 262
logging.log("WARNING: The path %s does not exist!", path)
```

## Root Cause

`logging.log()` signature is `logging.log(level, msg, *args)`. The caller passes the message string as `level`, which Python's logging module tries to coerce to `int`, raising:

```
TypeError: '<' not supported between instances of 'str' and 'int'
```

This silently kills any code path that checks for non-existent paths (e.g. plugin output directories).

## Expected behavior

The call should use `logging.warning()` directly:

```python
logging.warning("WARNING: The path %s does not exist!", path)
```

## Environment

- Python >= 3.11 (all versions affected)

saurabh4269:
## Description

Replace the incorrect `logging.log()` call in `owtf/utils/strings.py` with `logging.warning()`. `logging.log(level, msg, ...)` expects a numeric level as its first argument; the existing code passes the message string, which raises a `TypeError` at runtime.

## Related Issue

Closes #1404

## Motivation and Context

`paths_exist()` is called throughout the framework to validate plugin output directories and config paths. A `TypeError` here silently kills execution without logging the actual warning — the opposite of the intended behaviour.

## How Has This Been Tested?

**Reproducing the original failure:**

The old call `logging.log("WARNING: The path %s does not exist!", path)` raises `TypeError` at runtime because `logging.log()` expects a numeric level as its first argument:

```
$ python3 -c "import logging; logging.log('WARNING: msg %s', 'x')"
--- Logging error ---
TypeError: '<' not supported between instances of 'str' and 'int'
```

**Verifying the fix:**

The replacement `logging.warning("WARNING: The path %s does not exist!", path)` works correctly:

```
$ python3 -c "import logging; logging.warning('WARNING: The path %s does not exist!', '/nonexistent')"
WARNING:root:WARNING: The path /nonexistent does not exist!
```

**Test suite:**

```
$ pytest tests/
```

All pre-existing tests pass. No new test was added — this is a one-line API correction with no new code paths or branching logic to exercise.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/utils/strings.py
+++ b/owtf/utils/strings.py
@@ -259,7 +259,7 @@ def paths_exist(path_list):
     valid = True
     for path in path_list:
         if path and not os.path.exists(path):
-            logging.log("WARNING: The path %s does not exist!", path)
+            logging.warning("WARNING: The path %s does not exist!", path)
             valid = False
     return valid
 


viyatb (approved):
`logging.log()` requires a numeric level, so the current call can raise instead of reporting the missing path. `logging.warning()` is the correct one-line fix. This is ready.

PR #1407, opened 2026-03-31
fix: replace deprecated logging.warn() with logging.warning() in owtf/shell/

### LINKED ISSUE #1406: Bug: deprecated logging.warn() used in owtf/shell/base.py and owtf/shell/interactive.py
## Describe the bug

`logging.warn()` has been deprecated since Python 3.2 and emits a `DeprecationWarning` on every call. The following call sites still use it:

- `owtf/shell/base.py` line 204: `logging.warn(out.decode("utf-8"))`
- `owtf/shell/interactive.py` line 38: `logging.warn("ERROR - Communication channel closed - %s", abort_message)`
- `owtf/shell/interactive.py` line 56: `logging.warn("ERROR: read - The Communication channel is down!")`
- `owtf/shell/interactive.py` line 100: `logging.warn("ERROR: Run - The Communication Channel is down!")`

**Correction (2026-07-10):** this issue originally stated `logging.warn()` was "removed in Python 3.12" and would raise `AttributeError`. That's incorrect. Verified against CPython that it has never been removed from a released Python version: a removal was targeted for Python 3.13 but was reverted before that release shipped ([cpython/cpython#105376](https://github.com/python/cpython/issues/105376)). Empirically confirmed it still runs (with a `DeprecationWarning`) on Python 3.13.14.

Since the project targets `python >= 3.11`, replacing these calls with `logging.warning()` removes the deprecation-warning noise and keeps the code on the currently-recommended API, rather than fixing a crash.

Note: `owtf/shell/pexpect_sh.py` has the same issue at five additional call sites and was flagged in the issue.

## Expected behavior

All calls should use `logging.warning()`, which has been the correct method since Python 3.2.

## Environment

- Python >= 3.11 (deprecation warning on every affected call, not a crash)

saurabh4269:
## Description

Replace all `logging.warn()` calls in `owtf/shell/base.py` and `owtf/shell/interactive.py` with `logging.warning()`. `logging.warn()` has been deprecated since Python 3.2 and emits a `DeprecationWarning` on every call.

Changed call sites:
- `owtf/shell/base.py` line 204
- `owtf/shell/interactive.py` lines 38, 56, 100

## Related Issue

Closes #1406

## Motivation and Context

The project targets `python >= 3.11`. `logging.warn()` is deprecated and prints a `DeprecationWarning` on every call; `logging.warning()` is the correct, non-deprecated API.

**Correction (posted after initial submission):** this PR and its issue originally stated these calls raise `AttributeError: module 'logging' has no attribute 'warn'` on Python 3.12+. That was incorrect — verified against CPython that `logging.warn()` has never been removed from a released Python version (a removal targeted for Python 3.13 was reverted before that release shipped, see [cpython/cpython#105376](https://github.com/python/cpython/issues/105376)). This fix is still worth doing to eliminate the deprecation-warning noise and stay on the recommended API, just not for the crash reason originally stated. Apologies for not verifying that claim before opening.

## How Has This Been Tested?

Verified `grep -r 'logging\.warn(' owtf/shell/` returns no matches after the fix. Existing test suite passes.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/shell/base.py
+++ b/owtf/shell/base.py
@@ -201,7 +201,7 @@ class BaseShell(object):
         except KeyboardInterrupt:
             os.killpg(proc.pid, signal.SIGINT)
             out, err = proc.communicate()
-            logging.warn(out.decode("utf-8"))
+            logging.warning(out.decode("utf-8"))
             output += out.decode("utf-8")
             try:
                 os.killpg(
--- a/owtf/shell/interactive.py
+++ b/owtf/shell/interactive.py
@@ -35,7 +35,7 @@ class InteractiveShell(BaseShell):
         :rtype: `bool`
         """
         if not self.connection:
-            logging.warn("ERROR - Communication channel closed - %s", abort_message)
+            logging.warning("ERROR - Communication channel closed - %s", abort_message)
             return False
         return True
 
@@ -53,7 +53,7 @@ class InteractiveShell(BaseShell):
         try:
             output = recv_some(self.connection, time)
         except DisconnectException:
-            logging.warn("ERROR: read - The Communication channel is down!")
+            logging.warning("ERROR: read - The Communication channel is down!")
             return output  # End of communication channel
         logging.info(output)  # Show progress on screen
         return output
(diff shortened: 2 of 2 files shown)


viyatb (approved):
These are direct replacements for the deprecated `logging.warn()` alias, with no behavior change. The scope is clean and this looks ready.

PR #1409, opened 2026-03-31
fix: add errors='replace' to UTF-8 decoding in shell_exec_monitor()

### LINKED ISSUE #1408: Bug: UTF-8 decode without error handling in owtf/shell/base.py silently kills plugin execution
## Describe the bug

In `owtf/shell/base.py`, `shell_exec_monitor()` decodes subprocess output as UTF-8 with no error handling:

```python
# lines 196–205
logging.info(line.decode("utf-8").strip())
output += line.decode("utf-8")
# ...
logging.warn(out.decode("utf-8"))
output += out.decode("utf-8")
```

Security tools (nmap, sqlmap, nikto, etc.) routinely emit binary sequences, Latin-1 encoded strings, and other non-UTF-8 bytes in their output. Any such byte sequence causes an unhandled `UnicodeDecodeError` that propagates up through the plugin runner, silently terminating the plugin without saving captured output.

## Root Cause

`bytes.decode("utf-8")` raises `UnicodeDecodeError` on invalid byte sequences. The `errors=` parameter is not specified, so it defaults to `'strict'`.

## Expected behavior

Decoding should use `errors='replace'` (substitutes `U+FFFD` for invalid bytes) so that plugin output is always captured even when tools emit non-UTF-8 data:

```python
line.decode("utf-8", errors="replace")
```

## Impact

Plugin execution silently fails on any tool that produces binary or mixed-encoding output. No error is surfaced to the user and no output is saved.

saurabh4269:
## Description

Add `errors='replace'` to all `bytes.decode('utf-8')` calls inside `shell_exec_monitor()` in `owtf/shell/base.py`. Also updates the `logging.warn()` call on the same line to `logging.warning()` (removed in Python 3.12).

## Related Issue

Closes #1408

## Motivation and Context

Security tools (nmap, sqlmap, nikto, etc.) routinely emit binary sequences, Latin-1 bytes, and other non-UTF-8 data. With `errors='strict'` (the default), any such byte causes an unhandled `UnicodeDecodeError` that propagates out of the plugin runner, silently terminating execution and discarding all output collected up to that point. `errors='replace'` substitutes `U+FFFD` for invalid bytes, preserving the rest of the output.

## How Has This Been Tested?

Manually fed a subprocess that emits `b'\xff\xfe'` (invalid UTF-8) through the loop — output is now captured with replacement characters instead of raising. Existing test suite passes.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/shell/base.py
+++ b/owtf/shell/base.py
@@ -193,16 +193,16 @@ class BaseShell(object):
                 if not line:
                     break
                 logging.info(
-                    line.decode("utf-8").strip()
+                    line.decode("utf-8", errors="replace").strip()
                 )  # Show progress on the screen too!
                 output += line.decode(
-                    "utf-8"
+                    "utf-8", errors="replace"
                 )  # Save as much output as possible before a tool crashes! :)
         except KeyboardInterrupt:
             os.killpg(proc.pid, signal.SIGINT)
             out, err = proc.communicate()
-            logging.warn(out.decode("utf-8"))
-            output += out.decode("utf-8")
+            logging.warning(out.decode("utf-8", errors="replace"))
+            output += out.decode("utf-8", errors="replace")
             try:
                 os.killpg(
                     os.getpgid(proc.pid), signal.SIGTERM
--- a/tests/owtf/test_shell_utf8_decode.py
+++ b/tests/owtf/test_shell_utf8_decode.py
@@ -1,0 +1,172 @@
+"""
+Tests for the errors='replace' UTF-8 decode fix in owtf/shell/base.py.
+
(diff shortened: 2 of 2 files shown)


viyatb (changes requested):
The decode changes themselves are sensible and cover both normal output and the interrupt path. The 172-line test file only tests Python’s built-in `bytes.decode(errors="replace")`, though; it never calls `BaseShell.shell_exec_monitor()`, so it does not prove the production paths were changed. Please replace this with a small test using a fake process/stdout and call the real method, including one invalid byte sequence. That will give us useful regression coverage with much less test code.

PR #1411, opened 2026-03-31
fix: replace tail subprocess with pure Python in OutputFileHandler

### LINKED ISSUE #1410: Bug: OutputFileHandler shells out to 'tail' breaking portability and input validation
## Describe the bug

In `owtf/files/handlers.py`, `OutputFileHandler.get()` reads the last N lines of a file by calling the OS `tail` command via `subprocess`:

```python
# line 109
no_of_lines = self.get_argument("lines", default="-1")
if no_of_lines != "-1":
    data = subprocess.check_output(["tail", "-" + no_of_lines, abspath])
```

This has several problems:

1. **Portability** — `tail` is not available on Windows or in minimal Docker containers (e.g. distroless images). The `subprocess` call fails with `FileNotFoundError`.
2. **No integer validation** — `no_of_lines` is taken directly from the HTTP query string and concatenated into the command argument without being validated as an integer first. An attacker can supply an argument such as `-n` or any string that changes the `tail` invocation in unexpected ways.
3. **Unnecessary dependency** — reading the last N lines of a file is trivially achievable in pure Python using `collections.deque` or `file.seek()`, with no subprocess overhead.

## Expected behavior

Replace the `subprocess.check_output(["tail", ...])` call with a pure Python implementation that:
- Validates `lines` is a positive integer before use.
- Caps the value at a safe maximum to prevent memory exhaustion.
- Works on all platforms.

## Environment

- Any platform without `tail` (Windows, distroless Docker)
- Python >= 3.11

saurabh4269:
## Description

Replace the `subprocess.check_output(["tail", ...])` call in `OutputFileHandler.get()` with a pure Python implementation using `collections.deque`. Also removes the `subprocess` import and adds strict integer validation plus a maximum line cap on the query parameter.

## Related Issue

Closes #1410

## Motivation and Context

The existing code has three problems:
1. **Portability** — `tail` is unavailable on Windows and in minimal Docker containers; the call raises `FileNotFoundError`.
2. **No input validation** — the `lines` query param is passed directly to the command without being validated as an integer, allowing unexpected `tail` flag injection.
3. **Unnecessary subprocess** — last-N-lines is trivially achievable in pure Python.

The new implementation:
- Parses `lines` as `int` and returns HTTP 400 if it is not a valid integer.
- Clamps the value to `[1, 10000]` to prevent memory exhaustion on very large files.
- Uses `collections.deque(file, maxlen=n)` for an efficient single-pass tail.
- Removes the `subprocess` import entirely.

## How Has This Been Tested?

Manually verified against a 10 000-line file: `?lines=5` returns the last 5 lines, `?lines=abc` returns HTTP 400, and the full-file path is unaffected. Existing test suite passes.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/files/handlers.py
+++ b/owtf/files/handlers.py
@@ -2,13 +2,13 @@
 owtf.files.handlers
 ~~~~~~~~~~~~~~~~~~~
 """
+import collections
 import datetime
 import email.utils
 import hashlib
 import mimetypes
 import os
 import stat
-import subprocess
 import time
 
 import tornado
@@ -104,9 +104,15 @@ class OutputFileHandler(tornado.web.StaticFileHandler):
                     self.set_status(304)
                     return
 
-            no_of_lines = self.get_argument("lines", default="-1")
-            if no_of_lines != "-1":
-                data = subprocess.check_output(["tail", "-" + no_of_lines, abspath])
+            no_of_lines = self.get_argument("lines", default=None)
+            if no_of_lines is not None:
+                try:
+                    n = int(no_of_lines)
+                except ValueError:
+                    raise tornado.web.HTTPError(400)
+                n = max(1, min(n, 10000))
+                with open(abspath, "rb") as file:
+                    data = b"".join(collections.deque(file, maxlen=n))
             else:
                 with open(abspath, "rb") as file:
                     data = file.read()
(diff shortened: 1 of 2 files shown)


viyatb (changes requested):
Removing the `tail` subprocess is a good direction, but this changes an existing UI contract. The Workers page sends `lines=-1` for “All”; the old handler treated `-1` as the full file, while the new clamp turns it into 1 and shows only the last line. Please preserve `-1` as the all-lines sentinel (and check the existing 0-based line menu while you are there). The tests also call a copied `_read_tail()` helper rather than `OutputFileHandler`, so add a real handler-level case for `lines=-1` and one normal tail request.

PR #1415, opened 2026-03-31
fix: use socket.getaddrinfo() in get_ips_from_hostname() to return all IPs

### LINKED ISSUE #1414: Bug: get_ips_from_hostname() always returns a single IPv4 address due to gethostbyname()
## Describe the bug

In `owtf/utils/ip.py`, `get_ips_from_hostname()` is supposed to return all IP addresses for a hostname (its name, doc, and the caller in `target.py:676` storing results in `alternative_ips`). However it resolves the hostname using `socket.gethostbyname()`:

```python
ip = socket.gethostbyname(hostname)   # returns a single str, e.g. "93.184.216.34"
ipchunks = ip.strip().split("\n")    # always ["93.184.216.34"] — split on \n is dead code
return ipchunks
```

Two problems:

1. **Only one IP is ever returned.** `socket.gethostbyname()` returns one address string. The `split("\n")` is dead code — a bare IP string never contains a newline.
2. **IPv6-only or dual-stack hosts fail silently.** `socket.gethostbyname()` only resolves to IPv4. A hostname with an AAAA record but no A record raises `socket.gaierror`, which is caught and re-raised as `UnresolvableTargetException`, making the target unusable in IPv6-only environments.

The correct API is `socket.getaddrinfo()`, which returns all address families and all addresses for a hostname.

## Expected behavior

`get_ips_from_hostname("example.com")` should return a list of all resolved IPs (IPv4 and IPv6), de-duplicated, with the first entry matching what `get_ip_from_hostname()` returns.

## Impact

- `target_config["alternative_ips"]` always contains at most one IP, so the alternative-IP feature is effectively broken for multi-homed hosts.
- Targets with IPv6-only DNS records cannot be scanned.

saurabh4269:
## Description

Replace `socket.gethostbyname()` with `socket.getaddrinfo()` in `get_ips_from_hostname()` so the function actually returns all addresses (IPv4 and IPv6) for a hostname, and remove the dead `split('\n')` that never produced more than one element.

## Related Issue

Closes #1414

## Motivation and Context

`socket.gethostbyname()` returns one IPv4 address string. The existing `split('\n')` tried to extract multiple IPs from the result but a bare IP string contains no newlines, so the list always had exactly one entry. This meant:

- `target_config["alternative_ips"]` was always a single-element list, breaking the alternative-IP feature for multi-homed hosts.
- IPv6-only or dual-stack hostnames (AAAA record only) raised `UnresolvableTargetException` because `gethostbyname()` cannot resolve AAAA records.

`socket.getaddrinfo()` returns all address families and all addresses. The fix deduplicates results (getaddrinfo may return the same IP for multiple socket-type/protocol combinations) while preserving order, then returns the full list.

## How Has This Been Tested?

Added `tests/owtf/test_ip_get_ips_from_hostname.py` with 6 unit tests:
- Multi-homed host returns all IPs.
- Duplicate addresses from multiple socket types are de-duplicated.
- Bare IPv4/IPv6 literals bypass DNS and are returned directly.
- Unresolvable hostname raises the expected exception.
- Regression guard confirming the old single-IP limitation is fixed.

All 6 tests pass.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/utils/ip.py
+++ b/owtf/utils/ip.py
@@ -16,26 +16,31 @@ def get_ips_from_hostname(hostname):
 
     :param hostname: Target hostname
     :type hostname: `str`
-    :return: IP addresses of the target hostname as a list
+    :return: IP addresses of the target hostname as a list (IPv4 and IPv6)
     :rtype: `list`
     """
-    ip = ""
-    # IP validation based on @marcwickenden's pull request, thanks!
+    # If the hostname is already a bare IP address, return it directly.
     for sck in [socket.AF_INET, socket.AF_INET6]:
         try:
             socket.inet_pton(sck, hostname)
-            ip = hostname
-            break
+            return [hostname]
         except socket.error:
             continue
-    if not ip:
-        try:
-            ip = socket.gethostbyname(hostname)
-        except socket.gaierror:
-            raise UnresolvableTargetException("Unable to resolve: '{!s}'".format(hostname))
 
-    ipchunks = ip.strip().split("\n")
-    return ipchunks
+    try:
+        results = socket.getaddrinfo(hostname, None)
+    except socket.gaierror:
+        raise UnresolvableTargetException("Unable to resolve: '{!s}'".format(hostname))
+
+    # Extract the address string from each result tuple, preserving order and
+    # removing duplicates (getaddrinfo may return the same IP for several
+    # socket-type/protocol combinations).
+    seen = []
+    for r in results:
+        addr = r[4][0]
+        if addr not in seen:
+            seen.append(addr)
+    return seen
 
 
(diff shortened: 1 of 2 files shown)


viyatb (changes requested):
The `getaddrinfo()` approach and ordered de-duplication look reasonable. As with the timer/temp-dir PRs, though, these tests define and exercise a separate copy of the function and even a separate exception class. They would still pass if the production function were deleted. Please import `owtf.utils.ip.get_ips_from_hostname`, patch that module’s `socket` calls, and assert the real `UnresolvableTargetException`. It would also be useful to keep one mixed IPv4/IPv6 result case against the real function.

PR #1419, opened 2026-03-31
fix: use time.monotonic() for elapsed-time measurement in Timer

### LINKED ISSUE #1418: Bug: Timer uses wall-clock time for elapsed duration — susceptible to NTP and DST skew
## Describe the bug

In `owtf/utils/timer.py`, elapsed time for plugin and command execution is measured by subtracting two `datetime.datetime.now()` values:

```python
def start_timer(self, offset="0"):
    self.timers[offset]["start"] = self.get_current_date_time()  # datetime.now()

def get_elapsed_time(self, offset="0"):
    return datetime.datetime.now() - self.timers[offset]["start"]  # wall-clock diff
```

`datetime.datetime.now()` uses the system wall clock. This clock can jump forwards or backwards due to:
- NTP synchronisation (common on CI servers and long-running scans).
- DST transitions.
- Manual clock adjustments.

During a long scan (hours), a negative NTP step will make `datetime.now() - start` smaller than the true elapsed time; a positive step inflates it. This directly corrupts the `RunTime` field stored in the DB for every command and plugin.

`time.monotonic()` is the correct choice for measuring elapsed durations — it is guaranteed to never go backwards and is not affected by wall-clock adjustments (Python 3.3+).

## Fix

Store a `time.monotonic()` snapshot in `start_timer()` and use it in `get_elapsed_time()`. Keep `datetime.datetime.now()` for the display timestamps (`Start`, `End`) that go into the DB and the UI.

```python
import time

def start_timer(self, offset="0"):
    self.timers[offset]["start"] = self.get_current_date_time()
    self.timers[offset]["monotonic_start"] = time.monotonic()

def get_elapsed_time(self, offset="0"):
    elapsed_seconds = time.monotonic() - self.timers[offset]["monotonic_start"]
    return datetime.timedelta(seconds=elapsed_seconds)
```

The return type remains `datetime.timedelta`, so all callers (`get_elapsed_time_as_str()`, `get_time_as_str()`) are unaffected.

## Impact

Incorrect `RunTime` values in the database for any scan that runs across an NTP step or DST boundary.

saurabh4269:
## Description

Store a `time.monotonic()` snapshot in `start_timer()` and use it in `get_elapsed_time()` instead of subtracting two `datetime.datetime.now()` wall-clock values. The `datetime` timestamps used for display and database storage (`start`, `end`) are unchanged.

## Related Issue

Closes #1418

## Motivation and Context

`datetime.datetime.now()` reads the system wall clock, which can jump backwards during NTP synchronisation or DST transitions. During a long scan (hours), a negative NTP step would silently make `RunTime` in the DB smaller than the real elapsed time; a positive step inflates it. `time.monotonic()` is guaranteed never to go backwards and is unaffected by clock adjustments — it is the correct API for measuring elapsed durations (Python 3.3+, PEP 418).

The return type of `get_elapsed_time()` remains `datetime.timedelta`, so all callers (`get_elapsed_time_as_str()`, `get_time_as_str()`, `shell/base.py`, `plugin/runner.py`, etc.) are unaffected.

## How Has This Been Tested?

Added `tests/owtf/test_timer_monotonic.py` with five unit tests:
- Elapsed time reflects the monotonic diff (not wall clock).
- A backward wall-clock jump does not produce a negative elapsed time.
- `get_elapsed_time_as_str()` still returns a human-readable string.
- `start_timer()` still returns a `datetime` object.
- `monotonic_start` key is stored in the timer dict.

All five tests pass.

## Types of changes

### Checklist:

### DIFF
--- a/owtf/utils/timer.py
+++ b/owtf/utils/timer.py
@@ -8,6 +8,7 @@ human-readable form.
 """
 import datetime
 import math
+import time
 
 from owtf.settings import DATE_TIME_FORMAT
 
@@ -31,6 +32,7 @@ class Timer(object):
         """
         self.timers[offset] = {}
         self.timers[offset]["start"] = self.get_current_date_time()
+        self.timers[offset]["monotonic_start"] = time.monotonic()
         return self.timers[offset]["start"]
 
     def get_current_date_time_as_str(self):
@@ -51,14 +53,18 @@ class Timer(object):
         return datetime.datetime.now()
 
     def get_elapsed_time(self, offset="0"):
-        """Gets the time elapsed between now and start of the timer in Unix epoch
+        """Gets the time elapsed since the timer was started.
+
+        Uses time.monotonic() so the result is unaffected by NTP steps,
+        DST transitions, or manual clock adjustments.
 
         :param offset: Timer index
         :type offset: `str`
         :return: Time difference
-        :rtype: `datetime`
+        :rtype: `datetime.timedelta`
         """
-        return datetime.datetime.now() - self.timers[offset]["start"]
+        elapsed_seconds = time.monotonic() - self.timers[offset]["monotonic_start"]
+        return datetime.timedelta(seconds=elapsed_seconds)
 
     def get_time_as_str(self, timedelta):
         """Get the time difference as a human readable string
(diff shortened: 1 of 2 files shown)


viyatb (changes requested):
Using `time.monotonic()` for elapsed duration is the right change. The tests need to exercise the real `owtf.utils.timer.Timer`, though: this file defines a second copy of the class and only tests that copy, so the production implementation could regress while all five tests still pass. `owtf.utils.timer` can be imported without constructing a DB session; please import `Timer` from there and patch `owtf.utils.timer.time.monotonic` in the tests.

PR #1438, opened 2026-05-27
Fix Docker setup for Apple Silicon Macs and editable install

abhaybansal16:
**Problem**
- Setup fails on Apple Silicon (M1/M2/M3) without platform: linux/amd64
- Backend crashes on start because python setup.py install creates an egg 
  that breaks config path resolution (regression from #1356)
- pip cache purge fails with permission error due to Rosetta cache directory
- owtf command path was wrong in wait-for-it.sh entrypoint

**Changes**
- Add platform: linux/amd64 to all services in docker-compose.dev.yml
- Switch to pip install -e . (aligns with editable install from #1356)
- Fix pip cache purge to handle permission errors gracefully
- Fix entrypoint command to use absolute path /home/owtf/bin/owtf

**Testing**
Verified on macOS with Apple Silicon (M1) — make compose-safe runs successfully and OWTF starts correctly.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -10,6 +10,7 @@ The PRIMARY AUTHORS are:
 And here is an incomplete list of contributors who have submitted patches,
 reported bugs, provided ideas, helped answer questions, and generally made
 OWASP OWTF much better:
+ * Abhay Bansal @abhaybansal16
  * Adi Mutu @an_animal / @am06
  * Alessandro Fanio Gonzalez @alessandrofg
  * Alexandra Sandulescu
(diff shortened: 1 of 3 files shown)


abhaybansal16:
@viyatb @7a please check this and let me know is there's anything I should change.

viyatb (changes requested):
I followed the paths through the dev container and I think the editable install is going to break once Compose starts. The image installs from `/home/owtf`, then the dev volume mounts the repo root over `/home/owtf/owtf`, which moves the actual Python package one level deeper than the editable install expects. Also, forcing every service to `linux/amd64` makes Apple Silicon run under emulation rather than fixing it natively. Could you adjust the mount/install layout and add a Compose smoke test that starts the backend and imports `owtf`? I would only pin amd64 for a service that truly has no arm64 option.

PR #1441, opened 2026-06-01
Fix: Targets not showing in UI due to session_id bug

abhaybansal16:
**Problem**

Two bugs prevented targets from working correctly in the UI:

1. `Session.get_active()` returned a SQLAlchemy `Row` object (a tuple) instead of a plain integer. Any code passing this to a DB query got `psycopg2: can't adapt type 'Row'` — an internal server error.

2. `TargetConfigHandler.get()` never passed `session_id` when calling `get_target_config_dicts()`, so the query filtered against `session_id=None` and returned no targets even when they existed in the DB.

**Changes**

* `owtf/models/session.py`: extract integer from Row in `get_active()`
* `owtf/api/handlers/targets.py`: pass active `session_id` when fetching targets

**Testing**

Verified on Docker (Apple Silicon, M4) — targets can be added via the UI and appear correctly in the targets table.

### DIFF
--- a/owtf/api/handlers/targets.py
+++ b/owtf/api/handlers/targets.py
@@ -4,6 +4,7 @@ owtf.api.handlers.targets
 
 """
 from owtf.api.handlers.base import APIRequestHandler
+from owtf.models.session import Session as OWTFSession
 from owtf.lib import exceptions
 from owtf.lib.exceptions import InvalidTargetReference, APIError
 from owtf.managers.target import (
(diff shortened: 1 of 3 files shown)


abhaybansal16:
@viyatb @7a Please review this.

viyatb (changes requested):
The session ID fix here makes sense, but the output path change feels unrelated and changes where every scan writes data. Can you drop that part from this PR and keep this focused on the targets bug? Both backend CI jobs are failing on the current head as well, so please rebase and rerun them after the cleanup. Once this is just the session/target fix and the tests are green, it should be much easier to approve.

abhaybansal16:
Hi @viyatb Please review the latest changes. Thanks

viyatb (approved):
Rechecked after the scope cleanup and the update from develop. This is back to the two target/session fixes, the active-session value is now a plain integer, and all six fresh CI jobs pass. Looks good to me.

PR #1440, opened 2026-05-31
Feat: community plugin backend

piyush140104:
## Description

This PR integrates the Community Plugin Marketplace backend into OWTF and connects it with the previously implemented marketplace frontend.

The implementation includes:

1. Added required marketplace settings constants
2. Registered all six marketplace API routes
3. Added a placeholder admin_required decorator
4. Implemented AST-based plugin validation
5. Implemented sandboxed plugin execution
6. Added the UserPlugin database model
7. Added the community plugin manager
8. Added Tornado API handlers for marketplace operations

The user_plugins table is automatically created on startup using SQLAlchemy's  create_all(), so no manual migration is required.

## Motivation and Context

The marketplace frontend was already available, but there was no backend support for plugin upload, validation, listing, approval, rejection, or execution. This PR provides the backend foundation required to enable the complete marketplace workflow and makes the Community Plugin Marketplace functional end-to-end.

## Reviewers

@viyatb @7a 

## How Has This Been Tested?

**Unit Tests (60/60 Passing):**

```bash
pytest tests/test_plugin_validator.py -v
pytest tests/test_sandbox_runner.py -v
pytest tests/test_community_plugin_manager.py -v
```

**Manual API Testing:**

* Uploaded http_header_checker.py via POST /api/v1/community-plugins/upload

  * Verified AST validation passes successfully
  * Verified sandbox dry-run execution succeeds
  * Verified plugin is stored correctly

* Listed plugins via GET /api/v1/community-plugins/

  * Verified correct paginated response

* Retrieved a plugin via GET /api/v1/community-plugins/1/

  * Verified plugin details are returned correctly

* Executed a plugin via POST /api/v1/community-plugins/1/run

  * Tested against https://example.com
  * Verified execution and security header findings

* Rejected a plugin via POST /api/v1/community-plugins/5/reject

  * Verified status changes to rejected
  * Verified rejection reason is stored

* Approved a plugin via POST /api/v1/community-plugins/5/approve

  * Verified status changes back to approved

**Environment:**

* macOS
* Python 3.10
* PostgreSQL (Docker)

## Types of Changes

* [ ] Bug fix (non-breaking change which fixes an issue)
* [x] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to change)
* [ ] Other

### Checklist

* [x] My code follows the code style (modified PEP8) of this project.
* [ ] My change requires a change to the documentation.
* [ ] I have updated the documentation accordingly.

PR #1442, opened 2026-06-06
Feat: plugin manager integration

piyush140104:
## Description                                                                                                                                                                    
  Wires approved community plugins into OWTF's plugin pipeline so they appear alongside built-in plugins and run safely in an isolated sandbox process.                             
                                                                                                                                                                                    
  Changes made:                                                                                                                                                                                           
1. Added get_community_plugin_dicts to fetch approved plugins from user_plugins table in standard OWTF plugin dict format.                                                          
  2. Updated get_all_plugin_dicts to merge community plugins into every plugin listing response.                                                                                      
  3. Added PluginRunner._run_community_plugin to run community plugins via SandboxRunner in an isolated child process.                                                                
  4. Updated PluginRunner.run_plugin to route community vs built-in plugins correctly.                                                                                                
  5. Fixed docker-compose backend startup command and added PYTHONPATH env var.                                                                                                       
  6. Fixed deprecated image-webpack-loader options in webpack config.                                                                                                                 

 ## Motivation and Context
  Community plugins uploaded and approved through the marketplace had no way to actually run inside OWTF. This PR bridges that gap. Once a plugin is approved by an admin it shows up in the plugin list and can run against any target, safely sandboxed.

  ## Reviewers
  @7a @viyatb 

  ## How Has This Been Tested?
  - Ran python -m pytest tests/test_plugin_manager_integration.py -v, all 23 tests pass
  - GET /api/v1/plugins/ returns community plugin merged with built-ins (145 built-in + 1 community = 146 total)
  - POST /api/v1/community-plugins/1/run/ executed successfully via SandboxRunner and returned real findings
  - Frontend at http://localhost:8019 loads and shows the approved plugin in the marketplace

  ## Types of changes
  - [ ] Other

  ### Checklist:

### DIFF
--- a/owtf/managers/community_plugin.py
+++ b/owtf/managers/community_plugin.py
@@ -289,22 +289,28 @@ def get_community_plugin(session, plugin_id: int) -> Optional[Dict]:
     return plugin.to_dict()
 
 
-def run_community_plugin(session, plugin_id: int, target_url: str) -> Dict:
-    """Execute a community plugin via SandboxRunner.
+def test_run_community_plugin(session, plugin_id: int, target_url: str) -> Dict:
+    """Quick test run of a community plugin against a URL.
 
-    The plugin must be in 'approved' status.
-    Returns a dict describing the execution outcome.
+    This is just a smoke test so the uploader/reviewer can see the plugin
+    actually executes. It skips the normal scan pipeline on purpose
+    (no target manager, no worklist, no output saved). For a real scan,
+    use the plugin runner like the built-in plugins do.
+
+    Plugin must be approved.
     """
     plugin = session.query(UserPlugin).get(plugin_id)
     if plugin is None:
-        return {"success": False, "error": "Plugin not found"}
+        return {"success": False, "error": "Plugin not found", "non_persistent": True, "is_test_run": True}
     if plugin.approval_status != APPROVAL_APPROVED:
         return {
             "success": False,
             "error": "Plugin is not approved (status: {})".format(plugin.approval_status),
+            "non_persistent": True,
+            "is_test_run": True,
         }
     if not os.path.isfile(plugin.file_path):
-        return {"success": False, "error": "Plugin file missing from disk"}
+        return {"success": False, "error": "Plugin file missing from disk", "non_persistent": True, "is_test_run": True}
 
     result: SandboxResult = SandboxRunner.run(
         plugin_path=plugin.file_path,
@@ -312,7 +318,14 @@ def run_community_plugin(session, plugin_id: int, target_url: str) -> Dict:
         timeout=plugin.execution_timeout,
         memory_limit=plugin.memory_limit,
     )
-    return result.to_dict()
+    payload = result.to_dict()
+    payload["non_persistent"] = True
+    payload["is_test_run"] = True
+    return payload
+
+
+# Old name kept as an alias so existing callers still work for now.
+run_community_plugin = test_run_community_plugin
 
 
 def approve_community_plugin(session, plugin_id: int) -> Optional[Dict]:
--- a/owtf/managers/community_plugin.py
+++ b/owtf/managers/community_plugin.py
@@ -1,0 +1,354 @@
+"""
+owtf.managers.community_plugin
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+All database-level operations for community-uploaded plugins.
+
+This module owns the "write" path (upload → validate → store) and the
+"read" path (list, get, search, approve/reject).  The sandbox execution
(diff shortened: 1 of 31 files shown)


piyush140104:
Hey @7a @viyatb, when you have a moment can you please review this?

viyatb (commented):
Focused review comment for the plugin-manager integration. The auth, auto-approval, sandbox timeout, public serialization, and duplicate-upload-route comments posted on #1440 also apply to this branch because it carries that backend code; I kept the duplicate threads there and only added the integration-specific issue here.

owtf/plugin/runner.py
```diff
@@ -362,11 +368,51 @@ def run_plugin(self, plugin_dir, plugin, save_output=True):
...
+        )
+        if not result.success:
+            logging.warning(
+                "Community plugin %r did not complete cleanly: %s (exit_code=%s, timed_out=%s)",
+                plugin.get("name"),
+                result.error,
+                result.exit_code,
+                result.timed_out,
+            )
+        return [{"type": "community_output", "output": result.to_dict()}]
```
viyatb: A failed sandbox run is wrapped as normal plugin output and returned to `process_plugin()`, which then marks the run `Successful` and saves it through `save_plugin_output()`. For OWTF this is a scan-result integrity issue: a community plugin can crash or time out and still appear as a successful scan result. At minimum, keep failed sandbox results out of the successful-output path; the exact status label can be refined separately.

```suggestion
        output = [{"type": "community_output", "output": result.to_dict()}]
        if not result.success:
            logging.warning(
                "Community plugin %r did not complete cleanly: %s (exit_code=%s, timed_out=%s)",
                plugin.get("name"),
                result.error,
                result.exit_code,
                result.timed_out,
            )
            raise PluginAbortException(output)
        return output
```

owtf/managers/community_plugin.py
```diff
@@ -0,0 +1,354 @@
...
+    if not os.path.isfile(plugin.file_path):
+        return {"success": False, "error": "Plugin file missing from disk"}
+
+    result: SandboxResult = SandboxRunner.run(
+        plugin_path=plugin.file_path,
+        target_url=target_url,
+        timeout=plugin.execution_timeout,
+        memory_limit=plugin.memory_limit,
+    )
+    return result.to_dict()
```
viyatb: Since this PR adds community-plugin support to `PluginRunner`, the marketplace run path should avoid calling `SandboxRunner` directly if this is intended to be a real OWTF plugin execution. Direct sandbox calls bypass OWTF's normal target context, status handling, output persistence, ranking, and partial-output behavior. Either route this through the same `PluginRunner` / worklist path used by built-in plugins, or make this endpoint explicitly a non-persistent smoke test.

Suggested shape if this should be a real OWTF run:

```python
# Resolve or create the OWTF target for target_url first, then run through
# the normal plugin pipeline instead of invoking SandboxRunner directly.
target_manager.set_target(target_id)
plugin_dict = community_plugin_to_runner_dict(plugin)
output = runner.process_plugin(
    session=session,
    plugin_dir=COMMUNITY_PLUGINS_DIR,
    plugin=plugin_dict,
)
return {"success": output is not None, "output": output}
```

If this is meant to stay synchronous and non-persistent, I would rename it in the API/UI as a test run so users do not confuse it with OWTF scan execution.

viyatb (commented):
One more architecture-level comment on the marketplace run path: the PR now has both direct SandboxRunner execution and PluginRunner integration, so the API should make the intended execution path explicit.

owtf/plugin/runner.py
piyush140104: fixed. Now builds the output list first and raises PluginAbortException(output) when the sandbox result is unsuccessful so failed timed-out runs go through the partial output path in process_plugin instead of getting saved as Successful. Kept the warning log so we still see the failure reason in the server log.

owtf/managers/community_plugin.py
piyush140104: Renamed the endpoint and surfaced the intent for now. The path is /api/v1/community-plugins/{id}/test-run/ (old /run/ kept as an alias so the existing UI doesn't break), the manager function is test_run_community_plugin, the handler is CommunityPluginTestRunHandler and the response now carries non_persistent: true and is_test_run: true.The docstrings call out that this is a smoke test and not a real OWTF scan.

piyush140104:
Hey @viyatb, Please review the changes.

piyush140104:
Superseded by #1444, which consolidates the marketplace UI, backend, validator, admin review flow, DB upgrader, and trust-model docs into one PR after the Slack design 
  discussion with @viyatb. Closing in favour of that PR.

PR #1439, opened 2026-05-28
Feat: plugin marketplace UI

piyush140104:
## Description               
  Added the Plugin Marketplace page to the OWTF webapp. This is the frontend part of my GSoC project for community plugin support. The page lives at /marketplace and has three tabs:         
                                                                                                                      
 ### Browse: 
Shows all approved community plugins in a card grid. You can search by name and filter by group (web, network, auxiliary) and type (active, passive, etc).    
           
Screenshot 2026-05-28 at 3 42 17 PM
                                                                          
 ### Upload: 
A form where any logged in user can submit their own plugin for review. Has drag and drop for the .py file, fields for name, author, description, group, type, version and tags. The submit button stays disabled until the required fields are filled.

Screenshot 2026-05-28 at 3 43 16 PM

### Pending Review: 
Admin only tab where approved admins can approve or reject submitted plugins. Hidden from regular users.

Screenshot 2026-05-28 at 3 43 32 PM

 Also wired up the Redux reducer, saga, route at /marketplace and added the nav link in the top navbar.

  ## Motivation and Context
OWTF currently only ships with its own built in plugins. There is no way for the community to contribute new plugins or share scanning scripts with other users. This marketplace is the foundation for community plugin support, one of the main goals of my GSoC project.

  ## Reviewers
  @7a @viyatb 

  ## How Has This Been Tested?
  Tested manually by running the webpack dev server locally and going through each tab in the browser.

  Checked:
  - Browse tab renders without errors (shows empty state since no backend yet)
  - Search bar and filter dropdowns are visible and functional
  - Upload tab form validation works, submit stays disabled until name, author, description and file are filled
  - Drag and drop on the file dropzone works and shows the filename and size after selecting
  - Run plugin modal opens and closes correctly
  - /marketplace route works and the nav link shows up in the top bar
  - No console errors on page load

  ## Types of changes
  - [ ] Bug fix (non-breaking change which fixes an issue)
  - [x] New feature (non-breaking change which adds functionality)
  - [ ] Breaking change (fix or feature that would cause existing functionality to change)
  - [ ] Other

  ### Checklist:
  - [x] My code follows the code style (modified PEP8) of this project.
  - [ ] My change requires a change to the documentation.
  - [ ] I have updated the documentation accordingly.

PR #1444, opened 2026-06-25
Feat: community plugin marketplace

piyush140104:
## Description

This PR adds the full Community Plugin Marketplace to OWTF end to end. It replaces my three earlier PRs (#1439, #1440, #1442) and folds everything into one clean PR after the Slack discussion with @viyatb about dropping native sandboxing in favour of an admin review based trust model.

Here is what this PR does, in plain words.

### Security / Trust Model (summary)

Full doc: [`docs/community_plugin_trust_model.md`](docs/community_plugin_trust_model.md)

  - **Community plugins are NOT sandboxed at runtime.** An approved plugin runs with the same permissions as a built in OWTF plugin, through the normal `PluginRunner` path.
  - **AST validation is a best effort filter, not a security boundary.** It rejects the obvious stuff at upload time (`import subprocess`, `os.system`, `exec`, `eval`, dynamic imports, etc). It does not stop a
  determined attacker who smuggles behaviour through metaprogramming, decoded strings, or import time side effects.
  - **Admin review of the source is mandatory before publication.** Approval is the trust decision. Admins are expected to read the code, not just click a button. An `AST pass` means "worth showing to an admin", not
  "safe to publish".
  - **Two roles only.** Regular users can upload, browse approved plugins, and see their own uploads (with rejection reason). Admins can view source, approve, reject, delete, test-run, and read the review history. There is
   no default admin email. Admins are seeded via `OWTF_ADMIN_EMAILS` or managed with the `owtf-admin promote/demote/list` CLI.
  - **`file_path` never leaves the server.** Public and owner serializers strip it. There are pytest assertions that fail if it ever leaks.
  - **`test-run/` is admin only, rate limited (3 per minute per admin) and non persistent.** It exists so admins can smoke test a pending plugin without giving normal users a way to bypass the "admin reviews source" 
  guarantee.

  ### 1. Marketplace UI

  Adds a new page in the OWTF webapp at `/marketplace` with three tabs.

  - **Browse** shows all approved community plugins in a card grid with search, group filter, and type filter.
  - **Upload** is a drag and drop form where any logged in user can submit their own plugin (`.py` file plus name, description, author, group, type, version, tags).
  - **Pending Review** is an admin only tab where admins can see all uploaded plugins waiting for review, click View Code to read the source, then either Approve or Reject with a reason.

  ### 2. Backend

  Adds the backend for everything above.

  - A new `UserPlugin` SQLAlchemy model stores plugin metadata in the `user_plugins` table (auto created on startup, no migration needed).
  - A `community_plugin` manager handles upload, list, get, approve, reject, test run, and review history operations.
  - A set of Tornado handlers exposes all of this over the API.

  ### 3. AST based static analysis at upload time

  Before anything is stored, the plugin file is parsed and checked for blocked imports (`os`, `subprocess` with `shell=True`, `socket`, `ctypes`, etc.) and blocked built-ins (`eval`, `exec`, `compile`, etc.). If the 
  plugin uses anything dangerous it is rejected and never written to disk. If it passes AST it is saved as pending until an admin reviews it.

  ### 4. Plugin runner integration

  Wires approved community plugins into OWTF's normal plugin pipeline. The `get_all_plugin_dicts` helper now merges approved community plugins into the standard plugin list, so they show up next to built in plugins 
  everywhere (plugin listing API, worker, UI). When the runner sees `source="community"` it loads the plugin from its on disk path using OWTF's standard module loader and calls `run(PluginInfo)` exactly like a built 
  in plugin. There is no separate execution path anymore.

  ### 5. Admin enforcement

  Adds proper admin enforcement. There is now an `is_admin` boolean column on the `User` model. A real `admin_required` JWT decorator checks this flag (and returns 403 if the user is not an admin) and is applied to 
  the approve, reject, review history, delete, and test run endpoints. A `/api/v1/community-plugins/me/` endpoint returns the current user's profile so the UI can decide whether to show the admin tab and badge. The UI 
  fails closed by defaulting `isAdmin` to false and only flipping it to true after `/me/` returns `is_admin: true`.

  Both `@jwtauth` and `@admin_required` now raise `tornado.web.Finish` on auth failure, so the handler body cannot accidentally run after a failed auth check.

  ### 6. Admin bootstrap (no placeholder default)

  There is no hard coded default admin email anymore. Admins are seeded through the `OWTF_ADMIN_EMAILS` env var (comma separated list). Whoever signs up with one of those emails gets `is_admin = True` set 
  automatically on user creation. For everything after that, there is a new `owtf-admin` CLI:

  ```
  owtf-admin promote reviewer@example.com
  owtf-admin demote  reviewer@example.com
  owtf-admin list
  ```

  If `OWTF_ADMIN_EMAILS` is empty and nobody has been promoted, the marketplace has no admins and no plugin can ever be approved. That is intentional. A shipped default admin is a credential leak waiting to happen.

  ### 7. Reviewer tracking and review history

  The `UserPlugin` model stores `user_id` (the uploader), `reviewed_by_user_id` (who approved or rejected) and `reviewed_at` (when). A new admin only endpoint `GET /api/v1/community-plugins/{id}/audit/log/` returns 
  the chronological lifecycle of any plugin (uploaded, approved or rejected, with timestamps and reasons).

  ### 8. Rate limiting on test run

  `POST /api/v1/community-plugins/{id}/test-run/` is admin only and limited to 3 calls per minute per user, so nobody can hammer the box with repeated test runs. It goes through the normal `PluginRunner` path with 
  `save_output=False` so the result is shown inline without polluting scan results.

  ### 9. Idempotent DB upgrade on startup

  New installs get the schema from `Model.metadata.create_all` at startup. Existing installs that already have a `user_plugins` table from an earlier version of this branch pick up any new columns via 
  `owtf.db.upgrade.run_startup_upgrades`, which runs `ALTER TABLE ADD COLUMN` for every column the model expects but the table does not have. It is idempotent, safe to call on every boot, and does not touch existing 
  data. No Alembic migration is needed.

  ### 10. Sandbox removed

  Removed the old sandbox subprocess module (`owtf/plugin/sandbox.py`) and its test file entirely, along with every reference to `SandboxRunner` in the codebase. The trust boundary is now admin review of the source 
  code, which is the model @viyatb suggested.

  ## Related Issue

  Replaces #1439, #1440 and #1442 with one consolidated PR after the design feedback round on Slack. Those PRs have been closed as superseded.

  ## Motivation and Context

  OWTF currently ships with only its own built in plugins, so the community cannot contribute or share new scanning scripts. The marketplace is the foundation that lets that happen.

  The earlier three PRs tried to make this safe by running community plugins inside a native subprocess sandbox. After discussing with @viyatb on Slack, we agreed that native sandboxing is a headache to do correctly 
  (timeouts, process groups, resource limits, all on bare macOS or Linux without containers) and that the right model for OWTF is a trust based one. Admins read the source of every plugin before approval, AST checking
   blocks the obviously bad uploads, and approved plugins then run like any other plugin. This PR implements that decision.

  ## Reviewers

  @viyatb @7a

  ## How Has This Been Tested?

  ### Automated tests

  All green on macOS Python 3.13:

  ```bash
  pytest tests/test_plugin_validator.py            -q   # 27 passed
  pytest tests/test_community_plugin_manager.py    -q   # 22 passed
  pytest tests/test_plugin_manager_integration.py  -q   # 22 passed
  pytest tests/test_jwtauth_decorator.py           -q   # decorator control flow + admin gate
  pytest tests/test_community_plugin_handler_auth.py -q # per endpoint role checks
  pytest tests/test_db_upgrade.py                  -q   # idempotent ALTER upgrader
  pytest tests/test_cli_admin.py                   -q   # owtf-admin promote/demote/list
  ```

  Between them these cover filename sanitisation, AST validation rules, upload, list, get, search, approve, reject, delete, the community plugin to runner dict conversion, the merge into the built in plugin list, the 
  runner routing for community vs built in plugins, the JWT / admin decorator control flow, per endpoint role checks (`file_path` never leaks, non admins get 403 on admin endpoints), the startup DB upgrader, and the 
  admin CLI.

  ### Manual end to end

  Tested on a local OWTF stack (Postgres in Docker, webapp on `http://localhost:8019`):

  1. Registered a regular user `user@example.com` with password `Test@1234`. Only Browse and Upload tabs visible, no ADMIN badge. Uploaded a sample plugin from the Upload tab. AST validation passed, plugin stored as 
  pending, success toast shown.
  2. Registered `admin@owtf.org` (seeded via `OWTF_ADMIN_EMAILS`) with password `Admin@1234`. The account was auto promoted to admin on user creation. After login the ADMIN badge appeared and the Pending Review tab 
  became visible.
  3. Opened Pending Review. The plugin uploaded by the regular user was listed. Clicked View Code, the full source opened. Clicked Approve, confirmation dialog appeared, the plugin moved out of pending. Repeated with 
  a second plugin and Reject with a reason. The rejection reason was stored and the plugin no longer appeared in Browse.
  4. Logged back in as the regular user. The approved plugin was now visible in Browse alongside built in plugins. The rejected plugin was still visible in my own uploads with the rejection reason.
  5. As admin, clicked Test Run on the approved plugin, entered a URL, output was returned inline (and not saved into the scan history). Spammed the button: the 4th request in a minute returned a 429 with the rate 
  limit error message.
  6. Hit `GET /api/v1/community-plugins/1/audit/log/` with the admin token, got back the chronological events (uploaded, approved with reviewer id and timestamp).
  7. Hit the same review history endpoint with the regular user's token, got a 403 as expected. Same for `/approve/`, `/reject/`, `/delete/`, `/test-run/`, and `/source/`.
  8. Confirmed public list and get responses never include `file_path` (there are also pytest assertions for this).
  9. Used the `owtf-admin` CLI to promote a second reviewer, listed admins, then demoted them again. All three commands worked without touching the DB by hand.

  A short screen recording of the full upload to approve to run workflow is attached below.

  ## Screenshots (if appropriate)

  Full workflow recording: https://drive.google.com/file/d/1uDuG25Pt5tGe7J7C9KHHdvvRKnZ_UAj4/view?usp=sharing

  ## Types of changes

  - [ ] Other

  ### Checklist


piyush140104:
Hey @7a @viyatb, when you have a moment can you please review this.

 While you are looking at it, can you also help me with these few things I was confused about:
 
  1. Admin email default. I set ADMIN_EMAILS = ["admin@owtf.org"] as a placeholder so reviewers can just register that email and immediately get admin access without touching the database. Should I change the default to something else or remove the default entirely and require the OWTF_ADMIN_EMAILS env var to be set in production?
  2. Admin management UI. Right now the only ways to add or remove an admin are editing the env var or running a one line SQL update. Want me to add a small admin only UI (or a manage.py style CLI command)
  for promoting and demoting users in a follow up PR?

  Thanks for the review.

piyush140104:
Update: addressed step1 jwtauth control flow + access-control tests.

piyush140104:
Update: addressed step2

piyush140104:
Update: addressed step 3 (validator/runner contract alignment)

piyush140104:
Update: addressed steps 4,5,6

viyatb (changes requested):
Thanks for working through the earlier review.

1. Existing installations never receive `users.is_admin`. `User` now maps that non-null column, but `run_startup_upgrades()` only upgrades `user_plugins`. `create_all()` does not add columns to an existing `users` table, so upgraded deployments will hit a missing-column error on ordinary `User` queries and cannot use the admin bootstrap/CLI. Please add a `users.is_admin BOOLEAN NOT NULL DEFAULT FALSE` upgrade and a regression test starting from the pre-PR users schema.

2. Approved community plugins leak their absolute server `file_path` through the existing authenticated `GET /api/v1/plugins/` endpoint. `get_community_plugin_dicts()` adds `file_path`, and `PluginDataHandler` returns those dicts directly. This contradicts the trust-model guarantee that `file_path` never appears in an API response. Please separate the internal runner representation from the API representation (or strip internal fields before serialization) and cover the legacy plugin-list endpoint with a regression test.

3. A non-admin can fetch another user's pending or rejected plugin metadata by guessing its id. `CommunityPluginDetailHandler` calls `get_community_plugin()` without checking approval status or ownership, even though the documented role boundary says regular users can browse approved plugins and see only their own uploads. Please restrict non-admin detail reads to approved plugins or the uploader, and add handler coverage for pending/rejected plugins owned by another user.

piyush140104:
Hey @viyatb, thanks for the review. I pushed fixes for all three review points on `feat/community-plugin-marketplace`:

PR #1450, opened 2026-07-10
fix: replace deprecated logging.warn() and IOLoop.instance() in proxy startup

### LINKED ISSUE #1449: Bug: deprecated logging.warn() and IOLoop.instance() in owtf/proxy/main.py startup
## Expected Behavior
The proxy should start and stop cleanly on supported Python/Tornado versions, using non-deprecated logging and IOLoop APIs.

## Current Behavior
`owtf/proxy/main.py` uses two deprecated APIs in the proxy start/stop path:

1. `logging.warn()` (in `start_proxy()`) has been deprecated since Python 3.2 and emits a `DeprecationWarning` on every call. It has not been removed from any released Python version: a proposal to remove it was targeted for Python 3.13, but the removal was reverted before that release shipped (see [cpython#105376](https://github.com/python/cpython/issues/105376)). Replacing it with `logging.warning()` eliminates the deprecation-warning noise and keeps the code on the officially recommended API, ahead of any future removal attempt.
2. `tornado.ioloop.IOLoop.instance()` starts the loop in `pseudo_run()` and stops it in `clean_up()`. `IOLoop.instance()` is deprecated since Tornado 5; `IOLoop.current()` is the documented API for the loop bound to the current context, which is the correct reference in OWTF's forked multiprocessing model (`self.server.start(N)` forks N workers).

## Possible Solution
Replace `logging.warn(...)` with `logging.warning(...)`, and both `IOLoop.instance()` calls with `IOLoop.current()`.

## Steps to Reproduce
1. Run OWTF and trigger proxy startup (`owtf.proxy.main.start_proxy()`).
2. Observe the `DeprecationWarning` printed for the `logging.warn()` call (not a crash, just warning noise).
3. Neither call currently breaks startup on the pinned Tornado 6.2.0; this is forward-compatibility and warning-noise cleanup, not a crash fix.

## Context
Same class of bug fixed elsewhere in this codebase in `shell/base.py`/`shell/interactive.py` (#1407) and `managers/worker.py` (#1413) — this is the same pattern in the proxy module.

**Correction (posted after initial submission):** this issue originally stated `logging.warn()` was "removed entirely in Python 3.12" and would raise `AttributeError`, breaking proxy startup. That was incorrect. I verified against CPython that `logging.warn()` has never been removed from a released Python version — a removal attempt was reverted before Python 3.13 shipped, and I confirmed it still works (with a `DeprecationWarning`) on Python 3.13.14. Edited above to the accurate motivation.

## Your Environment
* Operating System and version: Kali Linux
* sudo access: N/A

saurabh4269:
## Description
`owtf/proxy/main.py` used two deprecated APIs in the proxy start/stop path. This replaces:
- `logging.warn()` → `logging.warning()` in `start_proxy()`.
- `tornado.ioloop.IOLoop.instance()` → `IOLoop.current()` at both call sites (start in `pseudo_run()`, stop in `clean_up()`).

Both are the same class of deprecated-API fix in the same file, grouped into one PR.

## Related Issue
Fixes #1449

## Motivation and Context
`logging.warn()` has been deprecated since Python 3.2 and emits a `DeprecationWarning` on every call. It has not been removed from any released Python version (a proposed removal targeted for Python 3.13 was reverted before that release shipped, see [cpython#105376](https://github.com/python/cpython/issues/105376)). Replacing it with `logging.warning()` removes the deprecation-warning noise and keeps the code on the officially recommended API. `IOLoop.instance()` is deprecated since Tornado 5; `IOLoop.current()` is the documented API for the loop bound to the current context, which is the correct reference in the proxy's forked multiprocessing model and stays on the recommended API (`instance()` remains present in Tornado through at least 6.5.5, deprecated since Tornado 5.0 as a plain alias for `current()`, not removed).

**Correction (posted after initial submission):** the PR and issue originally stated `logging.warn()` was "removed in Python 3.12" and would crash the proxy with `AttributeError`. That was incorrect. I verified against CPython that this method has never actually been removed from a released Python version; there was a removal attempt reverted before Python 3.13 shipped. A first attempt at this correction also overstated the `IOLoop.instance()` side ("removed in newer Tornado") without checking; verified against the installed Tornado 6.5.5 source that `instance()` remains present, deprecated since 5.0 as a plain alias for `current()`, not removed. Both fixes are still worth doing (deprecation-warning cleanup and staying on the recommended API), just not for the crash/removal reasons originally stated. Apologies for not verifying either claim before opening.

## Reviewers
@viyatb @7a

## How Has This Been Tested?
Mechanical API replacements, same pattern as #1407/#1413. Confirmed `owtf/proxy/main.py` still compiles and no deprecated calls remain. No behaviour change beyond the deprecation fixes under the current single-IOLoop-per-process model. No test added, consistent with how #1405/#1407 (identical mechanical renames) were handled.

## Types of changes

### Checklist:
- [x] Tests, linting, and type checks pass locally (`owtf/proxy/main.py` is outside the linted `PY_QUALITY_PATHS` set).
- [x] I have added myself to AUTHORS.md (included in this PR).

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -53,6 +53,7 @@ OWASP OWTF much better:
  * Robert Hansen @RSNake
  * Sachin Kamath @iamskamath - https://sachinwrites.xyz
  * Sandro Gauci @sandrogauci - http://www.enablesecurity.com
+ * Saurabh Gupta @saurabh4269 - https://saurabh4269.github.io/
  * Shadab Zafar
  * @sharad1126
  * Tapan Sharma
--- a/owtf/proxy/main.py
+++ b/owtf/proxy/main.py
@@ -187,7 +187,7 @@ class ProxyProcess(OWTFProcess):
             # To run any number of instances
             # "0" equals the number of cores present in a machine
             self.server.start(int(self.instances))
-            tornado.ioloop.IOLoop.instance().start()
+            tornado.ioloop.IOLoop.current().start()
         except BaseException:
             # Cleanup code
             self.clean_up()
@@ -199,7 +199,7 @@ class ProxyProcess(OWTFProcess):
         :rtype: None
         """
         self.server.stop()
-        tornado.ioloop.IOLoop.instance().stop()
+        tornado.ioloop.IOLoop.current().stop()
 
 
 def start_proxy():
@@ -221,7 +221,7 @@ def start_proxy():
             abort_framework("Inbound proxy address already in use")
         # If everything is fine.
         proxy_process = ProxyProcess()
-        logging.warn(
+        logging.warning(
             "Starting HTTP(s) proxy server at %s:%d",
             INBOUND_PROXY_IP,
             INBOUND_PROXY_PORT,


viyatb (approved):
This is a straightforward cleanup and the replacements match Tornado/Python’s current APIs. I also reran the nearby proxy unit tests: 13 pass. CI is green, so this looks ready to merge.

PR #1452, opened 2026-07-10
fix: populate response_interceptors in add_interceptor()/remove_interceptor()

### LINKED ISSUE #1451: Bug: InterceptorManager never populates response_interceptors, so intercept_response() is a no-op
## Expected Behavior
Every registered interceptor should run on both requests and responses, since every `BaseInterceptor` subclass is required to implement both `modify_request()` and `modify_response()` (both are abstractmethods).

## Current Behavior
`InterceptorManager.add_interceptor()` only appends/inserts into `self.request_interceptors`. `self.response_interceptors` is sorted in the same method but nothing ever inserts into it, so it stays permanently empty and `intercept_response()` is a complete no-op. The default `HeaderModifier`'s response-side header injection never actually applies to responses, despite being configured to.

## Possible Solution
Register each interceptor on both `request_interceptors` and `response_interceptors` in `add_interceptor()`. Since `response_interceptors` would then be used for the first time, `remove_interceptor()` (which currently only removes from `request_interceptors`) needs the matching fix, otherwise a "removed" interceptor would keep silently modifying every response.

## Steps to Reproduce
1. Create an `InterceptorManager` (or use the default one created by `ProxyHandler`).
2. Call `intercept_response()` on any response object.
3. Observe none of the default interceptors' `modify_response()` logic ever runs, and `response_interceptors` is always empty.

## Context
Found while reading the proxy interception pipeline in depth. Note: enabling the response chain does change live behaviour (the default `HeaderModifier` begins adding headers to real responses), so the fix should be paired with a fail-closed disclosure / optional config gate — see the PR.

## Your Environment
* sudo access: N/A

saurabh4269:
## Description
`add_interceptor()` now registers each interceptor on both `request_interceptors` and `response_interceptors`. `remove_interceptor()` now removes from both chains too, since leaving it fixed on only one side would replace an inert bug (interceptor never fires on responses) with an active one (interceptor keeps firing after being "removed").

## Related Issue
Fixes #1451

## Motivation and Context
Every `BaseInterceptor` subclass implements `modify_response()` as an abstract requirement, but `response_interceptors` was never populated, so `intercept_response()` was permanently a no-op. `get_interceptor()`/`get_interceptors()`/`enable_interceptor()`/`disable_interceptor()` are untouched — they operate on the interceptor object by identity, shared between both lists.

**Behaviour-change disclosure (fail-closed note):** `intercept_response()` is called in `finish_response()` on every proxied response, so on merge this fix does change live behaviour — the default `HeaderModifier` begins adding its `X-OWTF-Proxy` / `X-Intercepted` headers to real responses for the first time. I verified the other three default interceptors stay no-ops with default config (URLRewriter/BodyModifier early-return because a streamed `HTTPResponse.body` is empty and read-only; DelayInjector injects no latency with `response_delay=None`). No response body is modified by default. **If you'd prefer response interception gated behind a config flag and defaulted off (fail-closed for a security tool), I'm happy to add that — the add/remove lifecycle correctness fix stands on its own regardless.** I kept the diff minimal and did not touch `get_status()`/`get_interceptors()` (which still report only the request chain); can address in a follow-up.

## Reviewers
@viyatb @7a

## How Has This Been Tested?
`pytest tests/owtf/test_interceptor_manager_response_chain.py` — 4 tests: interceptors registered on both chains; `remove_interceptor()` removes from both; the default `HeaderModifier` applies its headers to a response via `intercept_response()`; and a bytes-body response is passed through unchanged (guards that the newly-live response chain does not raise or corrupt a real read-only bytes body). Confirmed the first three fail on the original code.

## Types of changes

### Checklist:
- [x] I have added myself to AUTHORS.md (in the proxy-startup deprecated-APIs PR of this series).

### DIFF
--- a/owtf/proxy/interceptor_manager.py
+++ b/owtf/proxy/interceptor_manager.py
@@ -65,8 +65,10 @@ class InterceptorManager:
         """
         if position is None:
             self.request_interceptors.append(interceptor)
+            self.response_interceptors.append(interceptor)
         else:
             self.request_interceptors.insert(position, interceptor)
+            self.response_interceptors.insert(position, interceptor)
 
         # Sort by priority
         self.request_interceptors.sort(key=lambda x: x.priority)
@@ -76,18 +78,26 @@ class InterceptorManager:
 
     def remove_interceptor(self, interceptor_id: str):
         """
-        Remove an interceptor by ID.
+        Remove an interceptor by ID from both the request and response chains.
 
         :param interceptor_id: ID of the interceptor to remove
         """
         # Find by name (using as ID for simplicity)
+        removed = None
         for i, interceptor in enumerate(self.request_interceptors):
             if interceptor.name == interceptor_id:
                 removed = self.request_interceptors.pop(i)
-                logger.info(f"Removed interceptor '{removed.name}'")
-                return
+                break
 
-        logger.warning(f"Interceptor with ID '{interceptor_id}' not found")
+        for i, interceptor in enumerate(self.response_interceptors):
+            if interceptor.name == interceptor_id:
+                self.response_interceptors.pop(i)
+                break
+
+        if removed:
+            logger.info(f"Removed interceptor '{removed.name}'")
+        else:
+            logger.warning(f"Interceptor with ID '{interceptor_id}' not found")
 
     def get_interceptor(self, interceptor_id: str) -> Optional[BaseInterceptor]:
         """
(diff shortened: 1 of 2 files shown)


viyatb (approved):
Reviewed the response-chain registration/removal change and its focused regression tests. The scope is narrow, the default response behavior is covered, and the targeted test suite passes locally.

PR #1443, opened 2026-06-08
feat: intelligent priority-based scheduler (Phase 1) [GSoC 2026]

abhaybansal16:
**GSoC 2026 — Phase 1: Query-time Plugin Priority Scheduler**

Mentors: @viyatb @7a

**Problem**

OWTF's worklist used pure FIFO ordering. A slow passive recon check could block a fast active vulnerability scan simply because it was inserted first.

**Solution**

Replace FIFO with priority-based ordering computed at query time via SQLAlchemy `case()` expressions. No schema changes, no new DB columns,no new dependencies — priority is purely a query ordering policy.

**Changes**

- `owtf/managers/scheduler.py` — new file with `PLUGIN_TYPE_PRIORITY`, `PLUGIN_RISK_PRIORITY`, and `plugin_priority_expr()`
- `owtf/managers/worklist.py` — `get_work_for_target()` joins Plugin and orders by priority expression instead of `Work.id`
- `tests/unit/managers/test_scheduler.py` — 9 unit tests passing

**Priority ordering**

Plugin type: active(40) > semi_passive(30) > passive(20) > grep(10) > external(0)

Risk bonus: SQLi/RCE/Command injection(+30), XSS/Auth bypass(+20), SSL/TLS(+10)

**Testing**

9 unit tests passing. OWTF starts and runs correctly with priority ordering active.

PR #1453, opened 2026-07-25
feat: dynamic worker scaling and batch work fetching

abhaybansal16:
**GSoC 2026 — Phase 2: Worker Optimization**
**Mentors - @viyatb @7a **

**Problem**

OWTF had no visibility into pending work and no infrastructure for batch task fetching. Workers always fetched one task at a time, limiting optimization opportunities.

**Solution**

Add pending work visibility and batch fetching infrastructure. Cap initial worker count at WORKER_MAX_PROCESSES. Infrastructure ready for future dynamic scaling optimization.

**Changes**

- `owtf/managers/worklist.py` — add `get_pending_count()` for work queue visibility and `get_work_batch()` for batch fetching
- `owtf/managers/worker.py` — cap initial workers at `WORKER_MAX_PROCESSES` in `spawn_workers()`
- `owtf/settings.py` — add worker configuration constants: `WORKER_BATCH_SIZE=5`, `WORKER_MAX_PROCESSES=8`, `WORKER_HIGH_WATER=50`, `WORKER_LOW_WATER=10`
- `tests/unit/managers/test_worker.py` — 5 unit tests for pending count and batch size validation

**Testing**

All tests passing. OWTF starts and runs correctly with worker optimization in place.


abhaybansal16:
@viyatb @7a Please review this and let me know if I need to make any changes.

viyatb (changes requested):
Thanks for the Phase 2 work. I found several blockers before this can merge:

1. Both required backend jobs fail in tests.functional.cli.test_empty_run (OWTF does not finish properly). Please fix the shutdown/scaling interaction and get the required matrix green.
2. spawn_workers() still creates PROCESS_PER_CORE * cpu_count workers at startup and never applies WORKER_MAX_PROCESSES. On machines where that value exceeds 8, the new maximum is already violated before dynamic scaling begins. Cap the initial process count too, and add behavior-level coverage.
3. test_get_work_batch_returns_correct_count imports get_work_batch but never calls it; it only verifies that seven inserted rows exist. Please test the real function, including batch-size limiting and ordering.
4. This PR is based on develop but contains the complete unmerged #1443 scheduler stack plus unrelated Docker/target fixes. Please retarget it to the Phase 1 branch while #1443 is open, or rebase onto develop after #1443 lands, so this PR contains only the Phase 2 delta.

There is also trailing whitespace in owtf/managers/worker.py and missing EOF newlines in the added scheduler/tests; please clean those up.

abhaybansal16:
Hi @viyatb Please review the changes. Thanks

viyatb (changes requested):
Re-reviewed the current head after the latest push. Phase 2 still is not implemented end to end: `WorkerManager.get_task()` and `manage_workers()` continue to fetch one item through `get_work_for_target()`, so `get_work_batch()` is never used; `WORKER_HIGH_WATER`, `WORKER_LOW_WATER`, and `get_pending_count` are imported but no scale-up/scale-down logic uses them. The branch only caps the initial worker count. Its focused test file also fails locally: 2 failed, 3 passed because both batch tests import nonexistent `owtf.db.session.session_manager`. Please integrate batch assignment and real watermark-driven scaling into the manager, replace those tests with an isolated SQLAlchemy fixture that exercises ordering/limits, and update/rebase now that #1443 is merged so this PR contains only Phase 2.

abhaybansal16:
@viyatb Can you review the changes. Thanks

viyatb (changes requested):
I found two blockers in the current batch flow. WORKER_BATCH_SIZE isn’t imported in worker.py, so manage_workers raises NameError on its first pass. More importantly, get_work_batch deletes the whole batch before the manager knows how many workers are idle; on a busy cycle, the unassigned rows are lost. Please only claim as many rows as there are idle slots and keep the rest queued. The scale-down branch currently only logs too, so either implement the drain or leave scaling out of this PR. A test with busy workers and extra queued jobs should verify that no work IDs disappear.

abhaybansal16:
Hi @viyatb Please review the changes

viyatb (changes requested):
Thanks, the missing import and batch limit are fixed. One queue lifecycle problem is still blocking this: get_work_batch marks rows inactive before a worker is actually ready, but those rows are never deleted or requeued. A draining worker also stays in self.workers and is counted as idle on later passes, even though the assignment loop always skips it. That can strand work and keep the queue count non-zero. Please claim only work that can be assigned, then delete or requeue it on failure, and remove and join drained workers. A manager-level two-cycle test would catch this.

abhaybansal16:
@viyatb Please review the changes

viyatb (changes requested):
The claimed-work tracking and drain removal are much better now. Two things still need fixing before this can merge. First, `LocalWorker` emits `failed` for an ordinary plugin exception, and `_resolve_worker_work()` requeues that row. A permanently broken plugin will therefore run forever instead of recording one failed run and letting the queue move on; only work interrupted by a dead/drained worker should be requeued. Please cover one `failed` result in the two-cycle test. Second, the new lifecycle test is not actually process-isolated: importing `WorkerManager` constructs the module-level singleton and starts real workers. I ran the file by itself on macOS and it fails during process spawn before the test body. Patch the singleton/spawn path or move the lifecycle logic behind an import-safe unit boundary so this test runs independently.

viyatb (approved):
This is in good shape now. I reran the manager and worker paths, including draining, interrupted work, and permanent plugin failures. Failed work is removed instead of being requeued forever, and the focused suite plus all six CI jobs are green.

PR #1454, opened 2026-07-31
feat: resilient plugin execution harness with timeout and retry

abhaybansal16:
**GSoC 2026 Phase 3 Part 1: Resilient Plugin Execution Harness**

**Mentors:** @viyatb @7a 

**Problem**

Plugins could hang indefinitely with no timeout mechanism, causing OWTF to freeze. Additionally, there was no way to distinguish between successful, timed-out, and failed plugin executions.

**Solution**

Implemented signal-based timeout enforcement with distinct outcome types and automatic retry logic on errors (but not timeouts).

**Changes**

* `owtf/plugin/harness.py` — `execute_with_timeout()` with TimeoutResult and ErrorResult classes
* `owtf/settings.py` — `PLUGIN_TIMEOUT=300` (5 minutes) and `PLUGIN_MAX_RETRIES=2`
* `owtf/plugin/runner.py` — Outcome type checking to properly classify execution results
* `tests/unit/plugin/test_harness.py` — 12 unit tests covering timeout, error, retry, and handler restoration
* `tests/unit/plugin/test_runner_status.py` — 3 runner-status classification tests

**Implementation**

**Timeout Enforcement**
- Uses SIGALRM on Unix platforms (Linux, macOS)
- Gracefully falls back on Windows/macOS where SIGALRM unavailable
- Previous SIGALRM handler properly saved and restored

**Distinct Outcomes**
- `TimeoutResult` — Plugin execution exceeded timeout
- `ErrorResult` — Plugin raised exception
- `dict` — Successful execution with output
- Runner classifies each outcome and persists appropriately

**Retry Logic**
- Retries on errors up to `PLUGIN_MAX_RETRIES` times (default 2)
- Does NOT retry on timeout (returns immediately)
- Logs retry attempts
- Returns ErrorResult after exhausting retries

**Testing**

9 comprehensive unit tests + 3 runner-status tests verifying:
- Timeout behavior
- Error handling
- Retry logic
- Handler restoration
- Outcome classification

**Result**

* Plugin execution is timeout-protected, preventing framework hangs
* Execution outcomes tracked distinctly (success/timeout/error)
* Failed plugins automatically retry, improving reliability
* Complete visibility into execution failures

PR #1455, opened 2026-07-31
feat: output deduplication for consistent reporting

abhaybansal16:
**GSoC 2026 Phase 3 Part 2: Output Deduplication**

**Mentors:** @viyatb @7a

**Problem**

Repeated scans produced duplicate findings, cluttering reports and wasting storage.

**Solution**

Database-backed deduplication using content fingerprinting with atomic constraints.

**Changes**

* `owtf/models/plugin_output.py` — added fingerprint column (String, nullable, unique, indexed)
* `owtf/plugin/normalizer.py` — `OutputDeduplicator` class for fingerprint computation
* `owtf/managers/poutput.py` — integrated dedup check before save, fingerprint persisted to database
* `tests/unit/plugin/test_normalizer_integration.py` — integration tests including save_plugin_output() called twice

**Implementation**

**Fingerprint Computation**
- SHA256 hash of: `plugin_code | target_id | output_json`
- Deterministic and consistent

**Duplicate Detection**
- Query database for existing fingerprint before save
- Skip save if found (duplicate)
- Persist fingerprint with output if new

**Database Upgrade**
- Fingerprint column auto-added via SQLAlchemy metadata.create_all()
- Works with existing databases

**Atomicity**
- PostgreSQL unique constraint prevents concurrent duplicates
- Safe for multiple workers

**Result**

* Duplicate findings eliminated from repeated scans
* Report clutter reduced
* Storage optimized

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -53,6 +53,7 @@ OWASP OWTF much better:
  * Robert Hansen @RSNake
  * Sachin Kamath @iamskamath - https://sachinwrites.xyz
  * Sandro Gauci @sandrogauci - http://www.enablesecurity.com
+ * Saurabh Gupta @saurabh4269 - https://saurabh4269.github.io/
  * Shadab Zafar
  * @sharad1126
  * Tapan Sharma
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -10,6 +10,7 @@ The PRIMARY AUTHORS are:
 And here is an incomplete list of contributors who have submitted patches,
 reported bugs, provided ideas, helped answer questions, and generally made
 OWASP OWTF much better:
+ * Abhay Bansal @abhaybansal16
  * Adi Mutu @an_animal / @am06
  * Alessandro Fanio Gonzalez @alessandrofg
  * Alexandra Sandulescu
--- a/docker/Dockerfile.backend
+++ b/docker/Dockerfile.backend
@@ -85,10 +85,10 @@ COPY --chown=owtf:owtf setup.py setup.cfg MANIFEST.in README.md ${HOME}/
 
 # Install OWTF using the recommended method (setup.py)
 RUN cd ${HOME} && \
-    python setup.py install
+    pip install -e .
 
 # Reduce cache footprint from pip to keep final image size smaller
-RUN pip cache purge && rm -rf ${HOME}/.cache
+RUN pip cache purge 2>/dev/null || true && rm -rf ${HOME}/.cache || true
 
 # Stage 3: Runtime image
 FROM base AS final_install
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -22,6 +22,8 @@ services:
     depends_on:
       - owtf-frontend
       - db
+    volumes:
+      - ~/.owtf:/home/owtf/.owtf
     networks:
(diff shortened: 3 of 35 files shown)


abhaybansal16:
@viyatb Please review this

viyatb (changes requested):
This needs changes before merge:

1. The new OutputDeduplicator is never imported or called by production code, so this PR does not prevent any duplicate PluginOutput rows. Please integrate it into the save path and add a database-backed integration test that proves a repeated save is skipped.
2. A process-local self.seen dictionary is not a reliable deduplication boundary for OWTF. Each worker has separate memory, it resets on restart, it misses rows already in the database, and concurrent workers can both accept the same fingerprint. Please make deduplication persistent and atomic at the database boundary, for example a stored fingerprint plus a unique constraint/upsert, with a migration and concurrency-safe tests.
3. Both required backend jobs fail in tests.functional.cli.test_empty_run because this branch includes #1453. This PR is based on develop but contains the full unmerged #1453/#1443 stack and unrelated Docker/target changes. Retarget to the immediate dependency branch while those PRs are open, or rebase onto develop after they land, so the dedup PR contains only its own delta.
4. The added files fail Ruff with whitespace/EOF errors; please format them.

The six helper tests only validate the in-memory dictionary and therefore do not cover the production requirement.

abhaybansal16:
@viyatb Please review the changes. Thanks

viyatb (changes requested):
Thanks for reworking this to use the database. I walked through the real save path and there are still a couple of things to fix before this can merge. `save_plugin_output()` calculates `fingerprint`, but it never passes that value into the new `PluginOutput`, so saved rows still have a null fingerprint and the next duplicate check will not find them. Existing databases also will not get the new column because there is no schema upgrade for it. The focused tests currently have 3 failures because they import a `session_manager` that does not exist. Could you wire the fingerprint into the saved row, add the database upgrade and a test that calls `save_plugin_output()` twice, and split the Phase 4 metrics/reporting files into their own PR? That will make this dedup change much easier to verify.

abhaybansal16:
Hi @viyatb Please review the changes. Thanks

viyatb (changes requested):
The fingerprint is written now, but existing databases never get the new column. I ran the new integration test against the current PostgreSQL schema and it fails with UndefinedColumn on plugin_outputs.fingerprint. Please add a startup/schema migration and test upgrading an old plugin_outputs table. The fingerprint should use plugin_key rather than plugin_code as well, since the same code can exist under different plugin types and those are separate outputs.

abhaybansal16:
Hi Viyat Please review the changes

viyatb (changes requested):
The migration and fingerprint scope look fixed, and the focused tests pass. There is still a race in the save path: two workers can both pass is_duplicate before either commits, then one hits the unique fingerprint index and fails instead of being treated as a duplicate. Please make the insert conflict-safe, or catch that unique violation and roll back, and test it with two sessions. This branch is also stacked on #1453 and #1454, which are still blocked.

viyatb (approved):
I reran this through the real save path with two independent database sessions racing on the same fingerprint. One output wins, the duplicate is safely ignored after rollback, and unrelated integrity errors still surface. The full fresh CI run is green, so this is ready to merge.

PR #1456, opened 2026-08-15
Community Plugin Marketplace: data model + DB upgrader

piyush140104:
## Description

  Foundation for the Community Plugin Marketplace series. First of six smaller PRs that together replace the giant #1444.

  This PR only touches the data layer:

  - New `UserPlugin` SQLAlchemy model backing a new `user_plugins` table. Holds the metadata for community-uploaded plugins (name, description, group, type, author, approval status, reviewer trail, resource
  limits, tags).
  - New `is_admin` column on the `users` table. The role is not enforced anywhere in this PR; that lands in the admin auth PR (next one in the series).
  - New `owtf.db.upgrade.run_startup_upgrades`. Runs `ALTER TABLE ADD COLUMN` for any column the ORM expects that a live table does not have. Idempotent, safe on every boot. Wired into `get_db_engine` right after
   `create_all`.
   
  No API endpoints, no handlers, no runner changes, no UI, no docs. Those follow in the next PRs so each one stays reviewable.
  
  ## Related Issue
  
  Splits the data layer out of #1444. Same underlying feature series, just smaller.
  
  ## Motivation and Context
  
  PR #1444 was too large for a comfortable review. As agreed with @viyatb, the plan is to break the marketplace into ~6 focused PRs so each piece can be reviewed on its own without holding up the rest.
  
  The DB upgrader exists because OWTF does not use Alembic. `create_all` builds missing tables but never adds missing columns to an existing one, so an operator upgrading an install would otherwise hit
  `UndefinedColumn` on the first query. The upgrader closes that gap without pulling in a full migration framework.
  
  ## Reviewers

  @viyatb @7a

  ## How Has This Been Tested?

  Local, Python 3.13, in-memory SQLite (no PostgreSQL needed).

  pytest tests/test_db_upgrade.py -q

  Result: 6 passed.

  Covers:

  - Fresh install: upgrader runs against an already up-to-date schema and does nothing.
  - Legacy `user_plugins` table (pre review-trail columns): upgrader adds every missing column.
  - Legacy `users` table (pre-marketplace, no `is_admin`): upgrader adds `is_admin` so any User query keeps working.
  - Idempotency: two consecutive `run_startup_upgrades` calls are a no-op the second time.
  - Missing table: upgrader silently skips instead of crashing.

  ## Types of changes

  - [ ] Other

  ### Checklist:

### DIFF
--- a/owtf/core.py
+++ b/owtf/core.py
@@ -4,13 +4,14 @@ owtf.core
 
 This is the command-line front-end in charge of processing arguments and call the framework.
 """
+
 from __future__ import print_function
 
-from copy import deepcopy
 import logging
 import os
 import signal
 import sys
+from copy import deepcopy
 
 from tornado.ioloop import IOLoop, PeriodicCallback
 
@@ -50,7 +51,7 @@ from owtf.settings import (
 from owtf.transactions.main import start_transaction_logger
 from owtf.utils.file import clean_temp_storage_dirs, create_temp_storage_dirs
 from owtf.utils.process import _signal_process
-from owtf.utils.signals import workers_finish, owtf_start
+from owtf.utils.signals import owtf_start, workers_finish
 
 __all__ = ["finish", "main"]
 
@@ -58,7 +59,8 @@ __all__ = ["finish", "main"]
 owtf_pid = None
 
 # Get a global DB connection instance
-from owtf.db.session import get_scoped_session
+from owtf.db.session import get_scoped_session  # noqa: E402
+from owtf.db.upgrade import run_startup_upgrades  # noqa: E402
 
 db = get_scoped_session()
 
@@ -77,14 +79,12 @@ def print_banner():
             @owtfp
         http://owtf.org
         Version: {0}
-        \033[0m""".format(
-            __version__
-        )
+        \033[0m""".format(__version__)
     )
 
 
 def get_plugins_from_arg(arg):
-    """ Returns a list of requested plugins and plugin groups
+    """Returns a list of requested plugins and plugin groups
 
     :param arg: Comma separated list of plugins
     :type arg: `str`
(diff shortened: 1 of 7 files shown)


viyatb (changes requested):
Thanks for splitting the model work out. One database blocker remains: `_add_missing_columns()` catches arbitrary DDL errors inside a single `engine.begin()` transaction and then continues. PostgreSQL marks that transaction aborted after the first error, so later ALTERs and the final commit still fail; the duplicate-column race described in the comment is not recovered. This is also invoked from `get_db_engine()`, so schema inspection/upgrades can run repeatedly rather than once at application startup. Please move this to an explicit one-time upgrade path and handle each expected duplicate-column race with PostgreSQL-safe DDL/savepoints or separate transactions. Add PostgreSQL coverage for an existing schema and a partially upgraded schema; the current SQLite tests cannot exercise the aborted-transaction behavior.

piyush140104:
hey @viyatb pushed a fix in 97c3b441. Each ALTER now runs in its own transaction so the postgres duplicate-column race doesn't kill the whole batch. Moved the upgrader out of get_db_engine into owtf.core.main so it fires once at startup and added a module-level guard. Regression tests against real postgres 16 for both a fully legacy schema and the partial-upgrade case, all green.

viyatb (changes requested):
Thanks for moving the upgrader out of get_db_engine and splitting the DDL into separate transactions. That fixes the original PostgreSQL problem. I still can’t approve this yet: the guard is set before the upgrade runs, and every DDL error is swallowed. A temporary DB error can leave the schema half-upgraded, and this process will never retry. Please set the guard only after a successful run and only ignore a confirmed duplicate-column race; other upgrade failures should stop startup. Also please revert the unrelated get_plugins_from_arg message change.

piyush140104:
hey @viyatb , pushed the fix guard now flips only after both upgrades succeed, except narrowed to only swallow duplicate-column race and reverted the unrelated get_plugins_from_arg change.

viyatb (changes requested):
Thanks, the latest upgrader changes address the transaction and retry concerns, and the focused suite now passes locally (10 passed, with 2 PostgreSQL-only tests skipped). One regression is still present in owtf/core.py: get_plugins_from_arg() now passes the message with a literal %s to usage(), so the actual plugin groups are no longer shown. The earlier comment says this unrelated change was reverted, but it is still in the current diff. Please restore the original formatting at line 97, and ideally keep the unrelated core.py formatting churn out of this data-layer PR. Once that is cleaned up, the database portion looks ready for another pass.

viyatb (dismissed):
Looks good now. I verified both fresh-database registration and the startup upgrade path, including repeat and failure behavior. The focused model/upgrader tests pass and all six CI jobs are green on this head.

viyatb (approved):
I checked this again on a fresh database and against the current develop stack. The user plugin table is registered on first startup, the upgrader remains idempotent for existing installs, and the focused schema tests plus all six CI jobs pass. This is ready to merge.

PR #1457, opened 2026-08-15
Community Plugin Marketplace: AST validator for uploads

piyush140104:
## Description

  Splits the AST validator out of #1444 into its own PR so it can be reviewed on its own. Independent of PR 1, no shared code.

  The validator runs at upload time, before the plugin file is written to disk, so obviously dangerous submissions are rejected instead of piling up in the admin review queue.

  What gets blocked:

  - Dangerous imports: `os`, `sys`, `socket`, `importlib`, `shutil`, `ctypes`, `multiprocessing`, `threading`, `signal`, `resource`, `pickle`, `shelve`, `marshal`, `pty`, `termios`, `tty`, `fcntl`, `builtins`,
  `gc`, `inspect`, `_thread`, `code`, `codeop`.
  - Built-ins: `eval`, `exec`, `compile`, `__import__`, `execfile`, `input`, `breakpoint`.
  - Attribute calls: `os.system`, `os.popen`, `os.exec*`, `os.fork`, `os.kill`, `os.remove`, `os.chmod`, `socket.socket`, and similar.
  - `subprocess.Popen` / `run` / `call` / `check_call` / `check_output` with `shell=True`.
  - `open(path, "w")` and every other write / append / exclusive mode. Read-mode `open()` is fine.
  - `async def` at any level.
  - Missing `DESCRIPTION` string constant or missing `run(PluginInfo)` entry point.

  What is allowed: `requests`, `subprocess.run([...])` without `shell=True`, `urllib`, `json`, and anything else that is not on the blocklist.

  This is a best-effort filter, not a sandbox. Admin review of the source before approval is the actual trust boundary; the validator just keeps the review queue clean.

  ## Related Issue

  Splits the AST validator out of #1444.

  ## Motivation and Context

  PR #1444 was too large to review comfortably. As agreed with @viyatb, the marketplace is being broken into six smaller PRs. This one is fully standalone (no manager, no handlers, no marketplace routes) so it
  can land independently of the other pieces.

  ## Reviewers

  @viyatb @7a

  ## How Has This Been Tested?

  Local, Python 3.13.

  pytest tests/test_plugin_validator.py -q

  Result: 29 passed.

  Covers:

  - Valid plugins: minimal template, `subprocess.run([...])` without shell, `requests`, `urllib`.
  - Blocked imports: `os`, `sys`, `socket`, `pickle`, `from os import ...`, `threading`.
  - Blocked calls: `eval`, `exec`, `compile`, `os.system`, `subprocess.run(shell=True)`.
  - File modes: `open(path, "w")` and `open(path, "ab")` rejected; `open(path, "r")` allowed.
  - Required structure: missing `DESCRIPTION`, missing `run`, both missing, syntax error.
  - `run(PluginInfo)` contract honoured; zero-arg `run()` warns and names the expected parameter.
  - `async def run` blocked.
  - `validate_file` and `validate_bytes` helper paths.

  ## Types of changes

  - [ ] Other

  ### Checklist:

### DIFF
--- a/owtf/plugin/validator.py
+++ b/owtf/plugin/validator.py
@@ -87,6 +87,16 @@ BLOCKED_ATTR_CALLS = frozenset(
 
 WRITE_MODES = frozenset({"w", "wb", "a", "ab", "x", "xb", "w+", "wb+", "a+", "ab+", "r+", "rb+"})
 
+SUBPROCESS_SHELL_CALLS = frozenset(
+    {
+        "subprocess.Popen",
+        "subprocess.run",
+        "subprocess.call",
+        "subprocess.check_call",
+        "subprocess.check_output",
+    }
+)
+
 
 @dataclass
 class ValidationResult:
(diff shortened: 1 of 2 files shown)


viyatb (changes requested):
The validator can be bypassed through an imported alias. This source currently passes with no violations: `from subprocess import run as process; DESCRIPTION = "x"; def run(PluginInfo): process("id", shell=True)`. The `shell=True` check only runs for `ast.Attribute` calls, while the aliased function is an `ast.Name`. The contract checks also accept a nested `run()` and any assignment named `DESCRIPTION`, including a non-string value, despite documenting top-level/module-level requirements. Please track import aliases (or block subprocess consistently), validate only `Module.body` for the entry-point contract, require a string DESCRIPTION, and add regression tests for these cases.

piyush140104:
hey @viyatb , fix is in 7e837f5d. The visitor tracks import aliases now, so aliased subprocess with shell=True gets caught the same as the direct call. Contract check only looks at tree.body so nested run() or nested DESCRIPTION don't count anymore and DESCRIPTION has to be a string literal.

viyatb (changes requested):
The import-alias fix works, and all 36 validator tests pass now. There are still two easy ways around the shell check: assign subprocess.run to a local name, or pass shell through a variable. I tried both and the validator accepted them. Please treat any shell value other than literal False as unsafe, and either follow simple assignment aliases or reject those indirect subprocess calls. A couple of focused tests for those two cases should close this out.

piyush140104:
hey @viyatb , pushed the fix shell= now only accepts  literal False (variables, truthy ints, expressions all flagged) and a new visit_Assign follows simple x = subprocess.run aliases so calls through x are caught the same as direct calls. 3 new tests added

viyatb (changes requested):
The subprocess fixes work, and all 39 validator tests pass locally. I found two same-class bypasses that are still accepted: `f = eval; f("1+1")` and `f = open; f("/tmp/x", "w")`. Alias resolution records both origins, but visit_Call checks the original function name for blocked built-ins and open(), so the resolved name is not used. A dynamic mode also passes: `mode = "w"; open("/tmp/x", mode)`. Please apply the resolved-name check to blocked built-ins and open(), and treat a supplied open mode as unsafe unless it is a literal read-only mode. Add focused tests for these cases. Since this validator is the upload security gate, these should be closed before merge.

viyatb (approved):
I exercised the validator against direct and aliased dangerous calls, file mode variants, the required run signature, and invalid UTF-8 input. The bypasses we found are covered now, failed uploads do not reach persistence, and all six fresh CI jobs pass. This is ready to merge.

PR #1458, opened 2026-08-15
 Community Plugin Marketplace: admin role, JWT decorators, owtf-admin CLI

piyush140104:
## Description

  Three pieces that together give the marketplace an admin role:

  - **`@jwtauth`** rewrite. The decorator now raises `tornado.web.Finish` on any auth failure, so the wrapped handler body cannot run after a 401. Fixes the "wrote 401 but the handler kept running" class of bug.
  Also handles single-word `Authorization: Bearer` headers safely.
  - **`@admin_required`** decorator. Layers on top of `@jwtauth`. First the token has to be valid (401 if not), then the resolved user has to be an admin either via `users.is_admin` or via the `OWTF_ADMIN_EMAILS`
   allow-list (403 if not). Raises `tornado.web.Finish` on either failure so the handler never runs.
  - **`owtf-admin` CLI.** New console script (`owtf-admin promote|demote|list`) that flips `users.is_admin` without touching the DB by hand. Registered in `pyproject.toml`.

  Supporting changes:
  - `add_user` promotes any email listed in `OWTF_ADMIN_EMAILS` on registration, so the first admin can log in without a manual DB update.
  - `settings.py` reads `OWTF_ADMIN_EMAILS` (comma-separated) at import time and logs a loud warning if it is empty, so operators notice the marketplace has no admin path configured.
  - `APIRequestHandler` gets two small helpers, `get_current_user_id()` and `get_current_user_obj()`, so admin handlers can look up the caller in one line.

  ## Related Issue

  Splits the admin auth layer out of #1444.

  ## Motivation and Context

  PR #1444 was too large to review comfortably. This piece is the auth foundation the marketplace endpoints (which land in the next PR) will build on. Kept independent of the marketplace itself so it can be 
  reviewed on its own merits.

  **Depends on PR (2/6) `[GSoC] Community Plugin Marketplace (1/6): data model + DB upgrader`** because `@admin_required` needs the `users.is_admin` column that PR 1 introduces. Once PR 1 merges, this branch will
   be rebased onto fresh `develop`.

  ## Reviewers

  @viyatb @7a

  ## How Has This Been Tested?

  Local, Python 3.13, in-memory SQLite where a DB is needed.

  pytest tests/test_jwtauth_decorator.py tests/test_cli_admin.py -q

  Result: 16 passed.

  Covers:

  - `@jwtauth`: missing Authorization header, malformed header, single-word header (regression), invalid JWT, valid JWT without a matching login row, valid JWT with a matching login row.
  - `@admin_required`: rejects non-admin (403), allows admin, rejects unauthenticated (401 from the inner `@jwtauth` layer).
  - `owtf-admin promote`: flips `is_admin`, idempotent on an existing admin, exits non-zero on unknown email.
  - `owtf-admin demote`: flips `is_admin`, idempotent on a non-admin.
  - `owtf-admin list`: shows DB admins plus the `ADMIN_EMAILS` allow-list.
  - Email lookup is case-insensitive.

  ## Types of changes

  - [ ] Other

  ### Checklist:

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -8,13 +8,12 @@ services:
     build:
       context: ..
       dockerfile: docker/Dockerfile.backend
-    command: ["/usr/bin/wait-for-it.sh", "db:5432", "--", "/home/owtf/bin/owtf"]
+    command: ["/usr/bin/wait-for-it.sh", "db:5432", "--", "owtf"]
     environment:
       - DOCKER=1
       - POSTGRES_USER=owtf_db_user
       - POSTGRES_PASSWORD=jgZKW33Q+HZk8rqylZxaPg1lbuNGHJhgzsq3gBKV32g=
       - POSTGRES_DB=owtf_db
-      - PYTHONPATH=/home/owtf/owtf
     ports:
       - 8008:8008
       - 8010:8010
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -8,12 +8,13 @@ services:
     build:
       context: ..
       dockerfile: docker/Dockerfile.backend
-    command: ["/usr/bin/wait-for-it.sh", "db:5432", "--", "owtf"]
+    command: ["/usr/bin/wait-for-it.sh", "db:5432", "--", "/home/owtf/bin/owtf"]
     environment:
       - DOCKER=1
       - POSTGRES_USER=owtf_db_user
       - POSTGRES_PASSWORD=jgZKW33Q+HZk8rqylZxaPg1lbuNGHJhgzsq3gBKV32g=
       - POSTGRES_DB=owtf_db
+      - PYTHONPATH=/home/owtf/owtf
     ports:
       - 8008:8008
       - 8010:8010
--- a/owtf/api/handlers/auth.py
+++ b/owtf/api/handlers/auth.py
@@ -3,42 +3,41 @@ owtf.api.handlers.auth
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
 """
-from sqlalchemy.sql.functions import user
-from owtf.models.user_login_token import UserLoginToken
-from owtf.api.handlers.base import APIRequestHandler
-from owtf.lib.exceptions import APIError
-from owtf.models.user import User
-from datetime import datetime, timedelta
-import bcrypt
-import json
-import jwt
+
+import logging
 import re
+import smtplib
+from datetime import datetime, timedelta
+from email.mime.multipart import MIMEMultipart
+from email.mime.text import MIMEText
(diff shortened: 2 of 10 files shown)


viyatb (changes requested):
The new demote command does not reliably remove admin access. `_set_admin(email, False)` clears `users.is_admin` and prints that the user is now regular, but `user_is_admin()` grants admin again on every request when the email remains in `OWTF_ADMIN_EMAILS`. That makes the CLI result misleading and leaves the account privileged. Please choose one source-of-truth behavior: use the allow-list only to seed the DB flag at registration/login, or have demote explicitly refuse/configure allow-listed users. Add a test that promotes, demotes, and then checks the same authorization path used by `admin_required`.

piyush140104:
hey @viyatb , fixed in 739af586. user_is_admin now reads only users.is_admin so demote actually takes effect. ADMIN_EMAILS still seeds admin at registration and now at login too, so pre-existing accounts on the allow-list get promoted automatically. Added a test that promotes then demotes an allow-listed user and checks user_is_admin returns False.

viyatb (changes requested):
The request-time admin check is using the database now, which is good. Demotion still won’t stick for an email in ADMIN_EMAILS, though: the next successful login sets is_admin back to true. Please keep the allowlist as a one-time seed instead of reapplying it on every login, and add a demote → login → still non-admin test. The Docker Compose changes also look unrelated to this PR and should come out.

piyush140104:
hey @viyatb , pushed the fix LogInHandler no longer reads ADMIN_EMAILS or writes is_admin, so demote sticks across logins, the allow-list stays as a one-time seed at User.add_user. Two new tests added.

viyatb (changes requested):
The demotion fix is present and the 19 focused unit tests pass locally. There is still a blocking issue in the decorator itself: `_reject()` raises `tornado.web.Finish` before Tornado enters the original `RequestHandler._execute()` method, so Tornado never catches it or finishes the response. I reproduced this with a real `AsyncHTTPTestCase`: a request without Authorization logs an uncaught `tornado.web.Finish` and the client gets `HTTPStreamClosedError` instead of the intended JSON 401. The fake-handler tests currently expect the exception to escape, so they hide the integration failure. Please either finish and return cleanly from the wrapper, or move the authorization check into a point covered by Tornado request handling, then add real HTTP tests for 401, 403, and the guarantee that the handler body does not run. This branch also needs to be rebased onto the final #1456 before it can merge.

viyatb (approved):
I retested the admin flow through the real JWT decorators and CLI paths. Unauthorized requests finish cleanly, promote/demote behavior is persistent and idempotent, and the admin allow-list is only used at account creation. All six fresh CI jobs pass on the current base. Ready to merge.

PR #1460, opened 2026-08-15
Community Plugin Marketplace: run approved plugins through the standard runner

piyush140104:
## Description

  Once a community plugin has been approved by an admin, it should behave exactly like a built-in plugin: same runner, same worker, same output plumbing. This PR is that bridge.

  **`owtf.managers.plugin`**

  - `get_community_plugin_dicts(session, for_api=False)`: reads approved rows from `user_plugins` and returns them in the built-in plugin dict shape (same `key`, `group`, `type`, `title`, `name`, `code`, `file`,
  `descrip`, `attr`, `min_time`). Runner path (`for_api=False`, default) also carries `source=community`, `file_path`, `execution_timeout`, `memory_limit`. API path (`for_api=True`) strips those internal fields.
  - `get_all_plugin_dicts(..., include_community=True, for_api=False)`: appends community entries to the built-in list and honours the same group/type filters. `include_community=False` keeps the old behaviour
  for callers that want built-ins only.

  **`owtf.plugin.runner.PluginRunner`**

  - `run_plugin` dispatches on `plugin["source"] == "community"`. Built-in path is unchanged.
  - New `_run_community_plugin` loads the plugin file from `plugin["file_path"]` via the standard module loader (`get_module`) and calls `run(plugin)`, exactly like a built-in plugin. Target URL falls back to
  whatever the plugin dict carries when `target_manager` is not configured (the test-run path).

  **`owtf.api.handlers.plugin.PluginDataHandler`**

  Every call site into `get_all_plugin_dicts` now passes `for_api=True` so `file_path` and other internal fields never leak through the legacy `/api/v1/plugins/` endpoint. Regression tests assert this.

  ## Related Issue

  Splits the runner integration out of #1444.

  ## Motivation and Context

  Without this bridge, approved community plugins would sit in the DB but never actually run under a scan. Keeping the merge in the manager (not the runner) means everywhere OWTF already asks the manager "what
  plugins exist?" now transparently sees the community ones.

  **Depends on PR 4 (community plugin backend), which brings in the `UserPlugin` model, manager, and settings.** Branch is stacked on PR 4 today; once PR 4 lands on `develop` this branch will be rebased and the
  diff will shrink to only PR 5's own changes.

  ## Reviewers

  @viyatb @7a

  ## How Has This Been Tested?

  Local, Python 3.13, in-memory SQLite.

  pytest tests/test_plugin_manager_integration.py -q

  Result: 26 passed.

  Covers:

  - `get_community_plugin_dicts` returns the right shape for approved plugins and skips pending/rejected ones.
  - Runner-only fields (`source`, `file_path`, `execution_timeout`, `memory_limit`) are present when `for_api=False` and absent when `for_api=True`.
  - `get_all_plugin_dicts` merges community plugins alongside built-ins and applies group/type filters.
  - `get_all_plugin_dicts(include_community=False)` excludes them.
  - `PluginRunner._run_community_plugin` loads the file via the module loader and invokes `run(plugin)`.
  - `PluginRunner.run_plugin` routes community vs built-in correctly based on `source`.
  - Legacy `/api/v1/plugins/` responses never contain `file_path` for community entries.

  ## Types of changes

  - [ ] Other

  ### Checklist:


viyatb (changes requested):
The runner dispatch works only when called directly; community plugins cannot pass through the real work queue. `get_community_plugin_dicts()` creates synthetic keys such as `passive@community_1`, but no corresponding row is added to the `plugins` table. `Work.plugin_key` is a foreign key to `plugins.key`, `add_work()` inserts that synthetic key, and worker retrieval dereferences `work_model.plugin`/joins `Plugin`. On PostgreSQL the insert violates the FK; without FK enforcement the worker query/backref still cannot retrieve the plugin. Please add a first-class persisted relationship or an explicit community-work queue path, then cover the full flow from `add_work()` through `get_work_for_target()` into `PluginRunner`. The current tests only call the manager/runner directly and miss this boundary. Also enforce the stored timeout and memory limit before treating those fields as execution controls.

piyush140104:
hey @viyatb, fixed in 76a32b6a. On approve I mirror the community plugin into the plugins table (source="community", file_path=...) so the worklist FK is real and the worker + runner dispatch work unchanged. Reject/delete unsync. Added SIGALRM timeout enforcement on the test-run path and an end-to-end integration test covering approve → add_work → get_work_for_target → Work.plugin. 109/109 tests pass.

viyatb (changes requested):
The plugins-table mirror fixes the work-queue foreign key, and the focused suite passes. One end-to-end gap remains: the mirror stores source and file_path, but not execution_timeout or memory_limit. The worker builds its plugin dict from that mirror, so real queued runs see timeout=0 and run without the configured bound. Please test from an actual Work row through _derive_work_dict and carry the limits into the runner path. If memory limiting is out of scope now, remove that setting instead of presenting it as enforced.

piyush140104:
hey @viyatb , pushed the fix mirror now carries execution_timeout onto the plugins-table row so Plugin.to_dict -> _derive_work_dict -> runner sees the configured bound instead of timeout=0. Added a new TestExecutionTimeoutFlowsFromWorkRowIntoRunner class that walks the real Work row path. Removed memory_limit entirely (column, settings, API kwarg, tests) since runtime memory limiting is out of scope.

viyatb (changes requested):
The execution_timeout now flows from the upload row through the plugins-table mirror, and the focused combined suite passes locally (84 passed, 2 PostgreSQL-only checks skipped). The remaining blocker is the mirror setting `Plugin.code = None`. The standard pipeline assumes code is a string identifier in several places. I reproduced two concrete failures: `show_plugin_types()` crashes with `TypeError: object of type NoneType has no len()`, and `get_all_plugin_dicts(session, {"code": ["Community"]})` mutates the requested name to `"None"` and returns no plugin. Output/report paths also store and index by plugin code, so a completed community run will not behave like a built-in plugin there. Please give mirrored plugins a stable identifier that the existing selection and reporting paths can use, or update those consumers consistently, then add an end-to-end test covering selection, execution, saved PluginOutput, and report retrieval. This branch must also be rebased onto the corrected earlier PRs before merge.

viyatb (approved):
I traced this from admin approval into the standard plugins table, worklist selection, worker dispatch, runner execution, output persistence, and report lookup. Community plugins now use the same timeout/error result path as built-ins, get stable unique plugin keys, and remain selectable by name. The 197-test combined backend run and all six fresh CI jobs pass. Ready to merge.

PR #1462, opened 2026-08-17
feat: Phase 4 - metrics, structured logging, and HTML reporting

abhaybansal16:
**GSoC 2026 Phase 4: Reporting, Metrics & Stability**

**Mentors:** @viyatb @7a

**Problem**

No visibility into plugin execution metrics, making it hard to identify bottlenecks or failures.

**Solution**

Added comprehensive metrics collection, structured logging, HTML reporting, and performance analysis.

**Changes**

* `owtf/plugin/metrics.py` — per-plugin execution metrics (success rate, runtime, failures)
* `owtf/utils/structured_logging.py` — structured logging format for all components
* `owtf/reports/html_report.py` — HTML report generation with per-plugin dashboard
* `owtf/performance.py` — performance optimizer with recommendations
* `owtf/plugin/normalizer.py` — integrated structured logging in deduplication checks
* `owtf/plugin/runner.py` — metrics recording in plugin execution
* `tests/benchmarks/benchmark_phase4.py` — performance benchmarking script

**Metrics Collected**

* Total runs per plugin
* Success/failure/abort counts
* Average/min/max runtime
* Success rate percentage
* Error tracking

**Reporting Features**

* Execution summary (total runs, success rate, failures)
* Per-plugin metrics table with tool attribution
* Performance benchmarking (before vs after)
* Performance optimization recommendations

**Structured Logging**

* Integrated into deduplication checks (normalizer.py)
* Logs duplicate detection events
* Consistent format across all components

**Result**

* Complete visibility into plugin execution
* Data-driven performance optimization
* Operational metrics for troubleshooting
* Integrated with Phases 1-3 for resilient, optimized execution

**Depends on:**
Phase 3 Part 2 (output deduplication) for complete functionality.

### DIFF
--- a/owtf/api/handlers/report.py
+++ b/owtf/api/handlers/report.py
@@ -8,16 +8,40 @@ from collections import defaultdict
 from time import gmtime, strftime
 
 from owtf.api.handlers.base import APIRequestHandler
-from owtf.constants import RANKS, MAPPINGS, SUPPORTED_MAPPINGS
+from owtf.api.handlers.jwtauth import jwtauth
+from owtf.constants import MAPPINGS, RANKS, SUPPORTED_MAPPINGS
 from owtf.lib import exceptions
 from owtf.lib.exceptions import APIError
 from owtf.managers.poutput import get_all_poutputs
 from owtf.managers.target import get_target_config_by_id
 from owtf.models.test_group import TestGroup
-from owtf.api.handlers.jwtauth import jwtauth
+
+__all__ = ["ReportExportHandler", "MetricsReportHandler"]
 
 
-__all__ = ["ReportExportHandler"]
+@jwtauth
+class MetricsReportHandler(APIRequestHandler):
+    """Return metrics aggregated from all persisted worker executions."""
+
+    SUPPORTED_METHODS = ["GET"]
+
+    def get(self):
+        from owtf.plugin.metrics import get_metrics
+        from owtf.reports.html_report import HTMLReportGenerator
+
+        metrics = get_metrics()
+        if self.get_argument("format", "json").lower() == "html":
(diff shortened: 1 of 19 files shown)


abhaybansal16:
@viyatb Please review this. Thanks

viyatb (changes requested):
There’s a blocker in the runner wiring: record_execution expects plugin_group, but process_plugin passes group. I reproduced the resulting TypeError; it happens in finally before output is saved. Please fix the keyword and add a test that actually runs process_plugin, not just the metrics class. After that, the metrics still need an end-to-end home: each worker process currently keeps its own in-memory instance, and the report generator isn’t connected to an API/CLI path that can aggregate those runs. This branch also inherits the #1455 schema-migration blocker.

abhaybansal16:
Hi Viyat Please review the changes.

viyatb (changes requested):
The persistence wiring is much better and the focused tests pass. I found two blockers in the end-to-end path. process_plugin assigns owtf_rank twice; the second unconditional call feeds timeout/error result objects into rank_plugin and can prevent partial output from being saved. Metrics are also grouped only by plugin_code, even though the same code is reused across active, passive, and grep plugins; I reproduced those runs collapsing into one row with the first type. Please remove the duplicate rank call and key persisted metrics by plugin_key, or by group, type, and code, with tests for both cases.

viyatb (approved):
I followed the runner all the way through status recording, output persistence, and reporting. Timeout/error results no longer get ranked, successful list output still does, and metrics now keep active/passive plugins with the same code separate in both memory and the database. The focused integration suite and all six fresh CI jobs pass. Ready to merge.

PR #1464, opened 2026-08-24
fix: broken lock-timeout handling in CacheHandler.load()

### LINKED ISSUE #1463: Bug: cache_handler.py references undefined name FileLockTimeoutException, not just a missing import
## Expected Behavior
CacheHandler.load() should handle a file-lock acquisition timeout gracefully (log and continue) when another process/worker holds the lock.

## Current Behavior
owtf/proxy/cache_handler.py:154 has `except FileLockTimeoutException:`, but FileLockTimeoutException does not exist anywhere in the codebase (owtf.lib.filelock's __all__ is ["FileLock"] and defines no such name). FileLock.acquire() actually raises the nested class FileLock.FileLockException on timeout. Any lock contention therefore raises NameError instead of being handled.

## Possible Solution
Catch FileLock.FileLockException instead (FileLock is already imported in this file). No new import needed.

## Steps to Reproduce
1. Instantiate a CacheHandler and call load() such that self.file_lock.acquire() raises FileLock.FileLockException (e.g. lock contention under concurrent proxy workers).
2. Observe NameError: name 'FileLockTimeoutException' is not defined instead of the intended debug-log-and-continue path.

## Context
Surfaces most under load / lock contention across concurrent proxy worker processes.

## Your Environment
* Installation method used: pip install -e . / make bootstrap
* sudo access: N/A

saurabh4269:
## Description
The except clause guarding FileLock acquisition in CacheHandler.load() was broken two ways, both fixed here:
1. It caught FileLockTimeoutException, a name that doesn't exist anywhere in the codebase. FileLock.acquire() actually raises the nested FileLock.FileLockException on timeout -- so catch that.
2. Its body logged `traceback.print_exc` (the function object, uncalled) as a format arg, which prints to stderr as a side effect and logs a useless function repr -- use traceback.format_exc() instead.

## Related Issue
Fixes #1463

## Motivation and Context
This wasn't a missing-import bug -- FileLockTimeoutException was never defined anywhere to import. The except clause was dead on arrival: any lock timeout raised NameError instead of being handled. Both fixes are in the same two-line except block, so they're handled together as "make the lock-timeout handler actually work."

## Reviewers
@viyatb @7a

## How Has This Been Tested?
Added tests/owtf/test_cache_handler_filelock_exception.py (2 tests): one confirms FileLock.acquire() raises FileLock.FileLockException on a pre-locked path; the other confirms CacheHandler.load() now returns None gracefully via a mocked timeout instead of raising NameError. Verified the second test fails with the original code (NameError) before this fix and passes after.

## Types of changes

### Checklist
- [x] Tests, linting, and type checks pass locally. New test added in tests/owtf/ (canonical test directory).

### DIFF
--- a/owtf/proxy/cache_handler.py
+++ b/owtf/proxy/cache_handler.py
@@ -11,7 +11,6 @@ import json
 import logging
 import os
 import re
-import traceback
 
 import tornado.httputil
 
@@ -105,6 +104,11 @@ class CacheHandler(object):
         :return:
         :rtype:
         """
+        file_lock = getattr(self, "file_lock", None)
+        if file_lock is None or not file_lock.locked():
+            logging.debug("Skipping cache write because this handler does not own the lock")
+            return
+
         try:
             response_body = self.request.response_buffer
             binary_response = False
@@ -152,7 +156,8 @@ class CacheHandler(object):
                 try:
                     self.file_lock.acquire()
                 except FileLock.FileLockException:
-                    logging.debug("Lock could not be acquired: %s", traceback.format_exc())
+                    logging.debug("Lock could not be acquired", exc_info=True)
+                    return None
                 # For handling race conditions
                 if os.path.isfile(self.file_path):
                     self.file_lock.release()
--- a/owtf/proxy/cache_handler.py
+++ b/owtf/proxy/cache_handler.py
@@ -151,8 +151,8 @@ class CacheHandler(object):
                 self.file_lock = FileLock(self.file_path)
                 try:
                     self.file_lock.acquire()
-                except FileLockTimeoutException:
-                    logging.debug("Lock could not be acquired %s", traceback.print_exc)
+                except FileLock.FileLockException:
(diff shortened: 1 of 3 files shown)


viyatb (changes requested):
Catching the right exception fixes the NameError, but the timeout path still continues as if this handler owns the lock. `load()` returns `None`, the proxy fetches and calls `dump()`, and `dump()` then writes the shared cache file and calls `release()` on a lock this instance never acquired. If the other request finished during the wait, line 158 can also unlink that other owner’s lock. Please return a distinct “not owner” outcome or retry/read the completed cache entry, and only write/release when `self.file_lock.locked()` is true. A two-handler test should verify that timing out cannot overwrite the cache or remove another handler’s lock.

viyatb (approved):
I fixed the remaining lock-ownership path and reran the focused tests. A handler that times out now returns without reading a partial cache entry, and dump() refuses to write or release unless that handler owns the lock. The two-handler regression test passes, and the full CI matrix is green.

PR #1466, opened 2026-08-25
fix: correct inverted SIGKILL escalation in _signal_process()

### LINKED ISSUE #1465: SIGKILL escalation is dead code in utils/process._signal_process() (inverted condition)
## Expected Behavior
_signal_process() sends a signal to a process tree, waits, then escalates to SIGKILL for any process that survived -- so a child that ignores the first signal is force-killed, not leaked.

## Current Behavior
The SIGKILL loop is guarded by `if not alive:` (owtf/utils/process.py), so it only runs when the survivor list is empty -- i.e. it never runs, and a surviving process is never killed. This is the same inverted condition as the WorkerManager._signal_children bug in #1412, but in a different function: _signal_process() is what WorkerManager.exit() uses to reap worker process trees, so the bug leaves orphaned workers on shutdown. This is one contributor to the web-mode "does not exit cleanly" behaviour that currently keeps the web/proxy functional tests excluded from CI.

## Possible Solution
Change `if not alive:` to `if alive:` so survivors are actually SIGKILLed.

## Steps to Reproduce
1. Have a worker process tree that does not terminate on the first signal.
2. Call WorkerManager.exit() (web-mode shutdown path) -> _signal_process().
3. Observe the survivor is never SIGKILLed (the escalation loop body executes zero times), leaving an orphan.

## Context
Twin of #1412. Surfaced while investigating the web-mode unclean-exit problem; this is the safely-shippable piece of that larger fix.

## Your Environment
* Installation method used: pip install -e . / make bootstrap
* sudo access: N/A

saurabh4269:
## Description
Changes `if not alive:` to `if alive:` in owtf/utils/process._signal_process(), so processes that survive the first signal are actually SIGKILLed instead of the escalation loop being dead code.

## Related Issue
Fixes #1465

## Motivation and Context
_signal_process() is the reaper WorkerManager.exit() uses to terminate worker process trees. With the inverted guard, survivors were never force-killed and leaked as orphans -- a contributor to the web-mode "does not exit cleanly" behaviour that keeps the web/proxy functional tests excluded from CI. This is the twin of the WorkerManager._signal_children bug in #1412 (fixed separately), in a different function.

## Reviewers
@viyatb @7a

## How Has This Been Tested?
`pytest tests/owtf/test_process_signal_sigkill.py` -- 3 tests using a mocked psutil: a survivor receives kill(); no survivors means no kill(); the first signal reaches both parent and children. Confirmed the survivor-kill test fails on the current code (dead loop) before this fix.

## Types of changes

### Checklist:
- [x] I have added myself to AUTHORS.md (done in the proxy-startup deprecated-APIs PR of this batch).

### DIFF
--- a/owtf/utils/process.py
+++ b/owtf/utils/process.py
@@ -62,8 +62,8 @@ def _signal_process(pid, psignal=signal.SIGINT):
     for pid in children:
         pid.send_signal(psignal)
     gone, alive = psutil.wait_procs(children, timeout=TIMEOUT)
-    if not alive:
-        # send SIGKILL
+    if alive:
+        # send SIGKILL to the processes that survived the first signal
         for pid in alive:
             logging.debug("Process %d survived SIGTERM; trying SIGKILL", pid)
             pid.kill()
--- a/tests/owtf/test_process_signal_sigkill.py
+++ b/tests/owtf/test_process_signal_sigkill.py
@@ -1,0 +1,56 @@
+"""Regression test for owtf.utils.process._signal_process().
+
+The SIGTERM -> SIGKILL escalation guarded its SIGKILL loop with `if not alive:`,
+so the loop only ran when the survivor list was empty -- i.e. it never ran, and
+processes that ignored the first signal were never force-killed (leaked as
+orphans). This is the same inverted-condition bug fixed in WorkerManager._signal_children
+(that one is a separate function / separate PR); this covers the twin in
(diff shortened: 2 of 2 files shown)


viyatb (approved):
Good catch. The condition was inverted, so the kill loop could only run when there were no surviving processes. The fix targets exactly the processes returned as alive, and the three focused regression tests pass locally. This looks ready.

